PR #1, opened 2015-04-28
Feature/sim 125 galsim noise

danielsf:
This ticket adds code to our GalSim interface that adds sky background and noise to the GalSim-generated images.


yoachim:
Looks good.  Approved.
**markdown**. I wonder if I can put emoticons in here.
💃💃💃

PR #6, opened 2015-05-21
updated docstrings in galSimNoiseAndBackground.py

danielsf:
This just updates the docstrings of the GalSim noise wrapper to be consistent with the new unit convention for readnoise, darkcurrent, and othernoise.


SimonKrughoff:
Good to go.

PR #9, opened 2015-06-03
Feature/sim 1101/photometric parameters

danielsf:
This pull request updates sims_catUtils to reflect the changes in the sims_photUtils pull request of the same name.

I also took the opportunity to clean up some of the docstrings in the GalSim interface, since that is where all of the changes were made.

### DIFF
--- a/python/lsst/sims/catUtils/galSimInterface/galSimCatalogs.py
+++ b/python/lsst/sims/catUtils/galSimInterface/galSimCatalogs.py
@@ -1,26 +1,3 @@
-"""
-The catalog classes in this file use the InstanceCatalog infrastructure to construct
-FITS images for each detector-filter combination on a simulated camera.  This is done by
-instantiating the class GalSimInterpreter.  This GalSimInterpreter is the class which
-actually generates the FITS images.  As the GalSim InstanceCatalogs are iterated over,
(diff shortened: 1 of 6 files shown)


python/lsst/sims/catUtils/galSimInterface/galSimCatalogs.py
```diff
@@ -81,9 +72,21 @@ class similar to drawPoinSource and drawSersic)
...
-    4) Telescope parameters such as exposure time, area, and gain are controled by the
-    GalSim InstanceCatalog member variables exposure_time (in s), effective_area (in cm^2),
-    and gain (in photons per ADU)
+    one should also define the following member variables:
+
+        componentList is a list of files ins banpass_directory containing the response
+        curves for the different components of the camera, e.g.
+        ['detector.dat', 'm1.dat', 'm2.dat', 'm3.dat', 'lens1.dat', 'lens2.dat', 'lens3.dat']
+
+        skyBandpassName is the name of the file in bandpass_directory that contains the sky
```
rhiannonlynne: Niggly comment about this docstring. I think if you call it 'atmospheric transmission curve' that it will be clearer what you actually want here (sky response curve doesn't actually mean anything obvious to me). So maybe update the rest of the string and change skyBandpassName to atmoBandpassName?

PR #11, opened 2015-06-15
Feature/sim 1182/customizable galsim

danielsf:
This pull request makes it easier to customize the CatSim-GalSim interface for use on other projects.

It adds member variables to the GalSimCatalog class that allow the user to point the catalog and different bandpass and SED data.  It alters the GalSimInterpreter to rely on instances of PhotometricParameters and ObservationMetaData to store observing conditions and telescope parameters.  It adds unit tests for these features, as well as for the code to add noise to images generated from a GalSimCatalog.

Note that there is a pull request of the same name in sims_catalogs_generation.


tests/testGalSimInterface.py
```diff
@@ -264,6 +344,73 @@ def catalogTester(self, catName=None, catalog=None, nameRoot=None):
...
+            cleanIm = afwImage.ImageF(cleanName).getArray()
+
+            totalVar = 0.0
+            totalMean = 0.0
+            ct = 0.0
+
+            self.assertEqual(cleanIm.shape[0], noisyIm.shape[0], msg='images not same shape')
+            self.assertEqual(cleanIm.shape[1], noisyIm.shape[1], msg='images not same shape')
+
+            for ix in range(cleanIm.shape[0]):
```
yoachim: Why loops? These are numpy arrays.

tests/testGalSimInterface.py
```diff
@@ -264,6 +344,73 @@ def catalogTester(self, catName=None, catalog=None, nameRoot=None):
...
+                    totalMean += mean
+                    ct += 1.0
+
+
+            totalVar = totalVar/ct
+            totalMean = totalMean/ct
+
+            if totalMean>=100.0:
+                countedImages += 1
+                self.assertTrue(numpy.abs(totalVar-1.0) < 0.05)
```
yoachim: Should I be scared that this is effectively testing that the noise is not a 3-sigma outlier? Does a seed need to be set somewhere in this method to make sure we don't get random outlier test failures?

yoachim:
ok, check those 2 comments and it looks good to merge.

tests/testGalSimInterface.py
danielsf: good point

tests/testGalSimInterface.py
danielsf: I have set the seed on the noise generating routine.

There is a random number generator internal to GalSim that is used to generate both the noiseless and the noisy images (for each object, photons are randomly drawn from the SED and shot at the image).  I do not see hot to access this generator to seed it.

PR #13, opened 2015-06-19
bugfix/SIM-1214/cosmologicalDimming

danielsf:
fixes unit tests to bring them in line with the same branch on sims_photUtils

### DIFF
--- a/python/lsst/sims/catUtils/utils/testUtils.py
+++ b/python/lsst/sims/catUtils/utils/testUtils.py
@@ -25,7 +25,7 @@ def calcADUwrapper(sedName=None, magNorm=None, redshift=None, internalAv=None, i
             a_int, b_int = sed.setupCCMab()
             sed.addCCMDust(a_int, b_int, A_v=internalAv, R_v=internalRv)
     
-    if redshift is not None and redshift!=0.0:
+    if redshift is not None and redshift != 0.0:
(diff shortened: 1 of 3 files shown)


python/lsst/sims/catUtils/utils/testUtils.py
```diff
@@ -26,7 +26,7 @@ def calcADUwrapper(sedName=None, magNorm=None, redshift=None, internalAv=None, i
             sed.addCCMDust(a_int, b_int, A_v=internalAv, R_v=internalRv)
     
     if redshift is not None and redshift!=0.0:
```
rbiswas4: - spaces sandwiching the != operator

rbiswas4:
Is there a comment somewhere about how to account for distances in these?

danielsf:
I'm not yet sure how we are supposed to correctly handle distances.  As I said (and am now writing to have a record of it), I think that the distance is folded into the magNorm.  If someone were using their own catalog, there is the CosmologyWrapper mixin which provides cosmologicalDistanceModulus as a column.  That would include both relativistic effects and distance (though, after our conversation with Lynne about what it means to redshift an Sed, I'm not even sure that is the correct thing to do).

We should probably design a rigorous set of tests to make sure that we know how to properly use all of the redshifting and renormalizing methods.  And then we should write down what we learn.

PR #19, opened 2015-08-13
Bugfix/sim 1315/registry bug

danielsf:
This pull request adds some unit tests to make sure that the changes implemented in the sims_catalogs_measures pull request of the same name have the expected effect on the Variability mixins.

I also change the applyVariability method in the main Variability mixin so that it throws a more helpful error message when you ask it for something it cannot do.


yoachim:
Approved!

PR #23, opened 2015-09-10
Feature/sim 1372/photometry redesign

danielsf:
This pull request updates the photometry mixins to make use of the new BandpassDict and SedList classes created in sims_photUtils.

@jbkalmbach -- again, please have a look if you have time.

### DIFF
--- a/python/lsst/sims/catUtils/mixins/PhotometryMixin.py
+++ b/python/lsst/sims/catUtils/mixins/PhotometryMixin.py
@@ -18,7 +18,7 @@ from lsst.sims.photUtils import Sed, Bandpass, LSSTdefaults, calcGamma, \
                                 loadTotalBandpassesFromFiles
 from lsst.sims.utils import defaultSpecMap
 from lsst.sims.catalogs.measures.instance import compound
-from lsst.sims.photUtils import CatSimSedList
+from lsst.sims.photUtils import SedList
 
 __all__ = ["PhotometryBase", "PhotometryGalaxies", "PhotometryStars"]
 
@@ -54,7 +54,7 @@ class PhotometryBase(object):
         a bandpass, which is to say that magnitudes[i][j] is the magnitude of the jth object in the
         bandpass characterized by self.bandpassDict.values()[i]
 
-        @param [in] bandpassDict is a CatSimBandpassDict characterizing the bandpasses being used
+        @param [in] bandpassDict is a BandpassDict characterizing the bandpasses being used
 
         @param [in] obs_metadata is the metadata of this observation (mostly desired because
         it will contain information about m5, the magnitude at which objects are detected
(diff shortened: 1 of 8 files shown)


python/lsst/sims/catUtils/mixins/PhotometryMixin.py
```diff
@@ -14,98 +14,202 @@
...
-    galaxies.  It assumes that we want LSST filters.
+    This class provides the basic infrastructure for photometry.
+    It can read in SEDs and bandpasses, apply extinction and redshift, and, given
+    an SED object it can calculate magnitudes.
+
+    In order to avoid duplication of work, the bandpasses, wavelength array, and phi array
+    are stored as instance variables once they are read in by either self.loadTotalBandpassesFromFiles()
+    or self.loadBandpassesFromFiles()
+
+    See the documentation of PhotometryHardware (from which this class inherits) for instructions how
```
rbiswas4: Is this still true?

danielsf: Wow.  No.  It is absolutely not true.  Good catch.

python/lsst/sims/catUtils/mixins/PhotometryMixin.py
```diff
@@ -14,98 +14,202 @@
...
+            raise RuntimeError("Passed %d magnitudes to " % magnitudes.shape[0] + \
+                                " PhotometryBase.calculatePhotometricUncertainty; " + \
+                                "needed %d " % len(bandpassDict))
 
-        @param [in] specFileMap is a mapping between the filenames in diskNames, bulgeNames, and agnNames
-        and the absolute locations of the corresponding files.  It is an instantiation of the class defined
-        in
+        #if we have not run this method before, calculate and cache the m5 and gamma parameter
+        #values (see eqn 5 of arXiv:0805.2366) for future use
+        m5Defaults = None
```
rbiswas4: Can we make this a keyword argument to the method with default None? Otherwise, maybe we can have a more LSST specific name rather than having something that sounds more general and always assigning LSST defaults to it.
I suppose that is that plan for later anyway.

danielsf: I don't think it should be a kwarg.  m5 is already stored in ObservationMetaData, and I don't want there to be too many places where the user has to set/pass those values.

I can change it so that it does not default to LSST values, though.

danielsf: This is actually a bigger change than I thought.  It causes some tangentially related unit tests to fail in interesting ways.  I think we should defer undoing the defaults to another issue.  I have opened SIM-1448 to address this during the next sprint.

rbiswas4: Unfortunately,  I cannot view this on github anymore. So, I am not entirely
sure of what you are referiing to:
Does this unit test failure have to do with m5 defaults or specFileMap ?

danielsf:
If I try to make m5 not default to LSSTdefaults, then testSetupMethods.py fails.  This because, when we try to automatically create a PhotometricCatalog using the methods defined in sims_catUtils/../utils/CatalogSetupFunctions.py, the created InstanceCatalogs rely on there being default m5 values to fill in the uncertainty values that aren't actually being requested.  This is a bug, since InstanceCatalogs created with the catalog setup functions are not supposed to calculate columns you don't actually want.  Untangling the bug is something I would rather undertake after this branch has been merged (since most of PhotometryBase as it is now was just copied and pasted from sims_photUtils).

PR #26, opened 2015-10-07
Feature/sim 1458/test ss mobj

danielsf:
I added a member variable ssmtable to SolarSystemObj indicating which of fatboy's table valued functions it was using.

I added daughter classes of SolarSystemObj to query the other table valued functions.

The script examples/checkSSMObj.py queries each of these tables and verifies that the resulting catalogs are not empty.


examples/checkSSMobj.py
```diff
@@ -0,0 +1,68 @@
...
+buildbot, Jenkins) to have a connection to fatboy.
+"""
+
+import numpy as np
+import os
+from lsst.utils import getPackageDir
+from lsst.sims.utils import ObservationMetaData
+from lsst.sims.catalogs.measures.instance import InstanceCatalog
+from lsst.sims.catUtils.baseCatalogModels import SolarSystemObj, CometObj, \
+                                                 MBAObj, NEOObj, \
```
rhiannonlynne: To match camelcase elsewhere, it would be acceptable to do MbaObj and NeoObj

python/lsst/sims/catUtils/baseCatalogModels/SsmModels.py
```diff
@@ -4,11 +4,28 @@
...
-__all__ = ["SolarSystemObj"]
+__all__ = ["SolarSystemObj", "CometObj", "NEOObj", "MBAObj", "MiscSolarSystemObj"]
 
 class SolarSystemObj(BaseCatalogObj):
+    """
+    This is the base CatalogDBObject from which all other Solar System CatalogDBObjects
+    derive.  It will query all of the Solar System object tables on fatboy.
+
+    Note: Solar System objects only exist for 5093.14 < mjd < 51923.0.  If you query outside
+    of that date range, you will get an empty catalog.
```
rhiannonlynne: typo in date range? I think it's 50093.14?

rhiannonlynne:
Generally looks good. 
I'm curious though - I can get a catalog written to diks using something like you have in the unit test. However, shouldn't I also be able to grab the column into in python using cat.column_by_name('raJ2000') or ('objid') (etc.)   but I can't. 

Traceback (most recent call last):
  File "catSimDiasources.py", line 36, in <module>
    print mySsmDb.column_by_name('objid')
  File "/Users/lynnej/lsstRepos/sims_catalogs_measures/python/lsst/sims/catalogs/measures/instance/InstanceCatalog.py", line 361, in column_by_name
    elif isinstance(self._current_chunk, _MimicRecordArray) or column_name in self._current_chunk.dtype.names:
AttributeError: 'NoneType' object has no attribute 'dtype'


Am I doing this wrong or or is there something else going on?

danielsf:
I'm actually not sure what happens if you call column_by_name() outside of the context of write_catalog()

my guess is that column_by_nam()e doesn't actually query the database.  Something else in write_catalog() has to do that first.

If you want to interact with the contents of a catalog as a numpy array in python, I suggest calling cat.iter_catalog() which will return an iterator over the rows of the catalog.

rhiannonlynne:
Ah thanks!
What I'd ideally like as the output would be a single file for more than
one visit (together with information about the visit metadata), so being
able to iterate over different catalogs and add the observation information
would be useful.

I realized there is not an obvious way to get the information about the
*orbits* of the objects. I guess this doesn't need to go into my output
file, and I probably have similar orbit files available elsewhere, but this
seems non-ideal, especially for people who might be coming in to query
fatboy and would need this information.  Any ideas?

rhiannonlynne:
Hmm - iter_catalog() isn't working how I'd expect either. I'll ask you
about it tomorrow.

danielsf:
Yeah.  Sorry.  After thinking about it more, iter_catalog() does not directly return a recarray.

iter_catalog() returns an iterator that allows you to iterate over rows of the catalog in python.  If you really want a numpy array, you're going to have to iterate over iter_catalog(), storing rows in a numpy array as you go.

I can look into adding a method that directly returns a recarray of results.  Any such method would probably only be limited to returning one chunk_size batch of rows at a time.

danielsf:
If you do not care about the order of rows in a joint catalog (a catalog comprising more than one visit), write_catalog has a kwarg 'write_mode' which, when set to 'a', will append whatever catalog it is writing to the specified file.  You will probably want to set write_header to False when doing that as well.

Regarding getting observation metadata into the catalogs, you can try overloading the write_header method in InstanceCatalog.

examples/checkSSMobj.py
danielsf: Do you prefer that?  I think I prefer it the way it is since "M", "B", and "A" all represent distinct words

rhiannonlynne: Good point, let's leave it as that then.

PR #27, opened 2015-10-12
Bugfix/sim 1557/test all objects

danielsf:
For a long time, all of the unit tests in testAllObjects.py have been marked as expected failures because they rely on connectivity to fatboy.  This is fine, but it means we do not catch failures in the test that are due to other problems.  This pull request makes it so that the unit tests only ignore failures that are, in fact, due to database connectivity.  Failures that come from other points in the code are treated as failed tests.


yoachim:
Looks good to me.  Merge away.

PR #29, opened 2015-10-15
Feature/sim 1504/ssm mixins

danielsf:
This should implement all of the solar system object mixins we talked about yesterday (October 14, 2015).

To add the focal plane, just make sure that your catalog class inherits from AstrometrySSM (which gives it access to raObserved and decObserved for solar system objects) and CameraCoords (which gives it access to chipName and pixel coordinates in the usual way).


python/lsst/sims/catUtils/mixins/PhotometryMixin.py
```diff
@@ -620,3 +621,133 @@ def get_lsst_magnitudes(self):
...
+        Method that actually does the work calculating magnitudes for solar system objects.
+
+        Because solar system objects have no dust extinction, this method works by loading
+        each unique Sed once, normalizing it, calculating its magnitudes in the desired
+        bandpasses, and then storing the normalizing magnitudes and the bandpass magnitudes
+        in a dict.  Magnitudes for subsequent objects with identical Seds will be calculated
+        by adding an offset to the magnitudes.  The offset is determined by comparing normalizing
+        magnitues.
+
+        @param [in] bandpassDict is an intantiation of BandpassDict representing the bandpasses
```
rhiannonlynne: 'instantiation'

python/lsst/sims/catUtils/mixins/PhotometryMixin.py
```diff
@@ -620,3 +621,133 @@ def get_lsst_magnitudes(self):
...
+        if not isinstance(self.obs_metadata.bandpass, str):
+            raise RuntimeError("dmagTrailing/dmagDetection calculation is confused. "
+                               "Your catalog's ObservationMetaData contains multiple "
+                               "bandpasses.  It is unclear what seeing value you mean. "
+                               "Re-create your catalog with only one bandpass.")
+
+        if not hasattr(self, 'photParams') or self.photParams is None:
+            raise RuntimeError("You cannot calculate dmagTrailing/dmagDetection. "
+                               "Your catalog does not have an associated PhotometricParameters "
+                               "member variable.  It is impossible to know what the exposure time is.")
```
rhiannonlynne: Hmm. Seems like it might be handy to either be able to change the seeing for PhotometricParameters or have exposure time available in obsMetadata (although I'd suggest implementing this as : a real default value is in photometricParameters and the obsMetadata *may* have an overriding value set -- this means that sims_photUtils still doesn't have to use obsMetadata, but that if someone is running through a series of pointings from an opsim run, they can easily change the visit exposure time for each exposure via the same interface they are updating the rest of the obs metadata).

python/lsst/sims/catUtils/mixins/PhotometryMixin.py
```diff
@@ -620,3 +621,133 @@ def get_lsst_magnitudes(self):
...
+        if self.obs_metadata.seeing is None:
+            raise RuntimeError("Cannot calculate dmagTraling/dmagDetection. "
+                               "Your catalog's ObservationMetaData does not "
+                               "specify seeing.")
+
+        if not isinstance(self.obs_metadata.bandpass, str):
+            raise RuntimeError("dmagTrailing/dmagDetection calculation is confused. "
+                               "Your catalog's ObservationMetaData contains multiple "
+                               "bandpasses.  It is unclear what seeing value you mean. "
+                               "Re-create your catalog with only one bandpass.")
```
rhiannonlynne: I'm not sure I understand this error .. if the bandpass is a str, and the seeing is a list, wouldn't that be fine? or if the seeing is a single value, couldn't it be applied to all bandpasses? (the trailing and detection loss itself don't care about the bandpass). 
I don't think this is going to be a problem for my use-case, but it doesn't seem to be required.

python/lsst/sims/catUtils/mixins/PhotometryMixin.py
```diff
@@ -620,3 +621,133 @@ def get_lsst_magnitudes(self):
...
+        if len(dradt)==0:
+            return numpy.zeros((2,0))
+
+        a_trail = 0.76
+        b_trail = 1.16
+        a_det = 0.42
+        b_det = 0.00
+        seeing = self.obs_metadata.seeing[self.obs_metadata.bandpass] # this will be in arcsec
+        texp = self.photParams.nexp*self.photParams.exptime
+        velocity = numpy.sqrt(numpy.power(numpy.degrees(dradt),2) + numpy.power(numpy.degrees(ddecdt),2))
```
rhiannonlynne: off topic: I want to get the velocity (rather than just separate velRa/velDec) into the output catalogs, so that it's accessible for MAF. I need to add another mixin method? (and can I put it into this PR?)

tests/testSSMmixins.py
```diff
@@ -0,0 +1,276 @@
...
+from lsst.sims.utils import ObservationMetaData
+from lsst.sims.catalogs.generation.db import fileDBObject
+from lsst.sims.catalogs.measures.instance import InstanceCatalog, compound
+from lsst.sims.catUtils.mixins import PhotometrySSM, AstrometrySSM
+from lsst.sims.photUtils import BandpassDict, SedList, PhotometricParameters
+from lsst.sims.coordUtils import _observedFromICRS
+
+class LSST_SSM_photCat(InstanceCatalog, PhotometrySSM):
+    column_outputs = ['id', 'lsst_u' ,'lsst_g', 'lsst_r', 'lsst_i', 'lsst_z', 'lsst_y']
+
```
rhiannonlynne: I like these unit tests as they're really helpful for me to understand how to use the new classes and mixins.

python/lsst/sims/catUtils/mixins/PhotometryMixin.py
danielsf: I'm leery of having any parameters stored both in ObservationMetaData and PhotometricParameters.  It could lead to ambiguity about which value takes precedence.

What if we did away with PhotometricParameters and stored all of its contents in ObservationMetaData?  I originally did not want to do this because I thought that ObservationMetaData had to contain a pointing RA and Dec.  I was wrong about that.  People just interested in sims_photUtils can instantiate an ObservationMetaData containing only gain, texp, nexp, etc. parameters, without having to specify a pointing.  This way, there will only be one class managing "state of the telescope" data.

I would want to open a new ticket to do this, as it would be a major, backwards-incompatible change touching most of the sims packages.

python/lsst/sims/catUtils/mixins/PhotometryMixin.py
danielsf: Rahul wanted to enable a single ObservationMetaData to contain multiple bandpasses (for use in simulating other surveys that can take images in more than one filter simultaneously).  I wanted to somehow catch that case and throw an error, since this mixin is only written to handle one seeing value.

You are right that this is the most conservative implementation of that check possible.  I can rewrite it so that, if ObservationMetaData contains multiple bandpasses but they all have the same seeing, the check still passes.

python/lsst/sims/catUtils/mixins/PhotometryMixin.py
danielsf: Yes.  I will add that method when I address your review.

python/lsst/sims/catUtils/mixins/PhotometryMixin.py
danielsf: Alternatively, we could move the definition of PhotometricParameters into sims_utils (which doesn't change much), and then have ObservationMetaData optionally carry around a PhotometricParameters instantiation inside it.  Then, catalogs wouldn't reference self.photParams, but self.obs_metadata.photParams, and users would be able to change all of the parameters in one place (the ObservationMetaData).

I think I prefer this option.

rbiswas4: I actually like the idea of having separate hardware/site attributes from observational parameters describing the pointing.  My impression was the ObservationMetaData and PhototmetricParameters  were set up to achieve this. In that sense, I would have thought that seeing belongs to ObservationMetaData, as does the exposure time, or is the feeling that exposure time is hard coded into the survey ?.

danielsf: Partially it was the (mistaken) belief that exposure time is a fixed constant for the survey.  Partially it is what they are used for.

PhotometricParameters was originally designed to hold parameters that the signal to noise methods in sims_photUtils needed.  ObservationMetaData was designed (originally) to hold information about the pointing that InstanceCatalog needed when querying fatboy.  Obviously, things have grown since then.  Anyway: the signal to noise methods require seeing and exposure time.  Ergo, seeing and exposure time went in PhotometricParameters.

If we were to move to a more physically-motivated system (hardware/site parameters going in PhotometricParameters; pointing parameters going in ObservationMetaData), we would be talking about a system where the signal to noise methods in sims_photUtils take both an ObservationMetaData and a PhotometricParameters (though, I guess even my proposed solution to this problem ends in that happening).

rbiswas4: Agreed. The question of splitting into physically motivated parameters matters only the plans include dealing with other telescopes/sites. If that scope is fixed, one can decide on whether to split these or lump them all in obsMetaData.

danielsf: I have opened ticket SIM-1569 to deal with this issue.

danielsf:
I have added a getter for skyVelocity to the AstrometrySSM mixin

PR #30, opened 2015-10-29
Bugfix/sim 1447/actual calculate columns

danielsf:
About a year ago, we introduced a modification to the photometry getters to ensure that an InstanceCatalog did not waste time calculating magnitudes that were not actually requested.

Unfortunately, our methods for calculating magnitude uncertainties broke that code.

This pull request patches that problem and also makes the framework for calculating magnitude uncertainties in an InstanceCatalog (hopefully) more easily extended.


python/lsst/sims/catUtils/utils/CatalogTestUtils.py
```diff
@@ -370,7 +370,7 @@ def get_cartoon_agn_mags(self):
         indices = [ii for ii, name in enumerate(self.get_cartoon_agn_mags._colnames) \
                    if name in self._actually_calculated_columns]
 
-        if len(indices)==6:
+        if len(indices)==5:
```
yoachim: Looks like some repeated code and a random magic number ==5.  Maybe make a _getIndices method with the 5 as a default kwarg?

yoachim:
yup, check one minor comment and this looks good to merge.

PR #31, opened 2015-11-03
Feature/add obs metadata

rhiannonlynne:
This seems to do what I need it to do. 
Example of use: 
https://github.com/testMOPS/mops/blob/master/catSimDiasources.py


rhiannonlynne:
Oops - something else is failing with the changes to the magnitude calculation.

python/lsst/sims/catUtils/mixins/ObsMetaMixin.py
```diff
@@ -0,0 +1,50 @@
+import numpy as np
+
+__all__ = ["ObsMetadataBase"]
+
+class ObsMetadataBase(object):
+    """
+    This class provides the InstanceCatalog with a way to write out parameters from its ObsMetadata object.
+    """
+    def _get_nobj(self):
+        return len(self.get_objId())
```
danielsf: should this be

return len(self.column_by_name('objId'))

It will be effectively the same thing, but a little more consistent with how getters are designed to work.

python/lsst/sims/catUtils/mixins/ObsMetaMixin.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def get_expMJD(self):
+        n_records = self._get_nobj()
+        mjd = [self.obs_metadata.mjd] * n_records
+        return np.array(mjd)
+
+    def get_rotSkyPos(self):
```
danielsf: add docstring explaining that this will be in degrees

python/lsst/sims/catUtils/mixins/ObsMetaMixin.py
```diff
@@ -0,0 +1,50 @@
...
+        n_records = self._get_nobj()
+        m5 = [self.obs_metadata.m5] * n_records
+        return np.array(m5)
+
+    def get_seeing(self):
+        n_records = self._get_nobj()
+        seeing = [self.obs_metadata.seeing[self.obs_metadata.bandpass]] * n_records
+        return np.array(seeing)
+
+    def get_fieldRA(self):
```
danielsf: add docstring explaining that this will be in degrees

python/lsst/sims/catUtils/mixins/ObsMetaMixin.py
```diff
@@ -0,0 +1,50 @@
...
+        n_records = self._get_nobj()
+        ra = [self.obs_metadata.unrefractedRA] * n_records
+        return np.array(ra)
+
+    def get_fieldDec(self):
```
danielsf: add docstring explaining that this will be in degrees

python/lsst/sims/catUtils/mixins/PhotometryMixin.py
```diff
@@ -148,6 +148,22 @@ def get_lsst_photometric_uncertainties(self):
...
+        @ param [in] magFilter is the magnitude of the object in the given filter.
+
+        @ param [in] obs_metadata is the observation metadata for a given visit (to determine m5).
+
+        @ param [out] visibility (0/1).
+        """
+        if len(magFilter) == 0:
+            return numpy.array([])
+        completeness = 1.0 / (1 + numpy.exp((magFilter - m5)/sigma))
+        probability = numpy.random.random_sample(len(magFilter))
```
danielsf: The catalogs won't be reproducible if we do this (numpy.random will be seeded from the system time each time the code is run).

Of course... even if we control the seed of numpy.random, we cannot guarantee that SQL will return objects in the same order.

I guess the best thing to do is just add a warning to the docstring explaining that the output of this getter is not going to be repeatable, even for nominally identical catalogs.

python/lsst/sims/catUtils/mixins/PhotometryMixin.py
```diff
@@ -148,6 +148,22 @@ def get_lsst_photometric_uncertainties(self):
                                                  'lsstBandpassDict')
 
 
+    def calculateVisibility(self, magFilter, m5, sigma=0.12):
+        """
+        Calculate the probability of detecting a particular source.
```
danielsf: This doesn't actually return the probability of an object being visible.  It returns a truth value (i.e. was it seen or wasn't it).

Should this method be split into two (one that calculates the probability that an object is visible; one that determines whether or not the object was seen) so that users have the option of outputting catalogs with a probability column?

python/lsst/sims/catUtils/mixins/PhotometryMixin.py
```diff
@@ -148,6 +148,22 @@ def get_lsst_photometric_uncertainties(self):
...
+
+        @ param [in] magFilter is the magnitude of the object in the given filter.
+
+        @ param [in] obs_metadata is the observation metadata for a given visit (to determine m5).
+
```
danielsf: You forgot to include a definition of sigma in the docstring.

python/lsst/sims/catUtils/mixins/PhotometryMixin.py
```diff
@@ -705,13 +721,56 @@ def get_lsst_magnitudes(self):
...
         if not hasattr(self, 'lsstBandpassDict'):
             self.lsstBandpassDict = BandpassDict.loadTotalBandpassesFromFiles()
 
         return self._magnitudeGetter(self.lsstBandpassDict, self.get_lsst_magnitudes._colnames)
 
 
+    def get_magFilter(self):
+        """
+        Generate the magnitude in the filter of the observation.
+        """
```
danielsf: Can you not just do:

magName = 'lsst_' + self.obs_metadata.bandpass
return self.column_by_name(magName)

Then, all of the work of loading the bandpassDict will be done by the compound getter for lsst_[u,g,r,i,z,y].

python/lsst/sims/catUtils/mixins/PhotometryMixin.py
```diff
@@ -705,13 +721,56 @@ def get_lsst_magnitudes(self):
...
+        # Get m5 for the visit
+        m5 = self.obs_metadata.m5[self.obs_metadata.bandpass]
+        # Adjust the magnitude of the source for the trailing losses.
+        dmagSNR = self.column_by_name('dmagTrailing')
+        magObj = (magFilter - dmagSNR).reshape((1, len(magFilter)))
+        if len(magObj) == 0:
+            snr = []
+        else:
+            snr, gamma = calcSNR_m5(magObj, [bandpass], [m5], self.photParams)
+            snr = snr.reshape((len(magFilter), 1))
```
danielsf: You should just be able to do 

return snr[0]

to get the numpy array corresponding to the magnitude you want

python/lsst/sims/catUtils/mixins/PhotometryMixin.py
```diff
@@ -705,13 +721,56 @@ def get_lsst_magnitudes(self):
...
+        Generate the magnitude in the filter of the observation.
+        """
+        if not hasattr(self, 'lsstBandpassDict'):
+            self.lsstBandpassDict = BandpassDict.loadTotalBandpassesFromFiles()
+        mags = self._magnitudeGetter(self.lsstBandpassDict, self.lsstBandpassDict.keys())
+        mag_keys = self.lsstBandpassDict.keys()
+        magFilter = mags[mag_keys.index(self.obs_metadata.bandpass)]
+        return magFilter
+
+    def get_SNR(self):
```
danielsf: should this method be renamed get_magSNR?

I can imagine adding a compound getter get_SNR that gets SNR for all of the bandpasses in the near future.

python/lsst/sims/catUtils/mixins/PhotometryMixin.py
```diff
@@ -705,13 +721,56 @@ def get_lsst_magnitudes(self):
...
+        if len(magObj) == 0:
+            snr = []
+        else:
+            snr, gamma = calcSNR_m5(magObj, [bandpass], [m5], self.photParams)
+            snr = snr.reshape((len(magFilter), 1))
+        return snr
+
+    def get_visibility(self):
+        """
+        Calculate the probability of detecting this particular source.
```
danielsf: Same comments as on the docstring for calculateVisibility

python/lsst/sims/catUtils/mixins/PhotometryMixin.py
rhiannonlynne: That's a good point. We could make the SQL return everything in order if we sorted by objId? But would that make things too slow? 
I would like to make it optionally at least potentially repeatable -- what do you do for setting random seeds in other instance catalogs?

python/lsst/sims/catUtils/mixins/PhotometryMixin.py
rhiannonlynne: I'm not sure that the probability column would ever be useful. I will update the doc string to reflect what it does return more accurately.

python/lsst/sims/catUtils/mixins/PhotometryMixin.py
danielsf: We don't have any instance catalogs that use random seeds, yet.  @rbiswas4 and @jbkalmbach are trying to figure out how to deal with this with supernovae and AGN, respectively.  No solution has been reached.  The best idea we've had is: re-seed numpy.random with objId for each object and just draw one random number from the resulting distribution.  Of course each supernova/agn just needs a few random numbers representing the parameters characterizing the light curve of the object.  Presumably, you need one random number for each permutation of objId + MJD (to account for the randomness involved in viewing one object on multiple occasions).

rhiannonlynne:
Addressed review comments and expanded documentation. All works now.

python/lsst/sims/catUtils/mixins/PhotometryMixin.py
rhiannonlynne: Okay. 
If I just want it to be generally reproducible, I just need to add the option to specify a seed here, and then somewhere I need to specify the seed and pass it in. 
If I am going to say "it's reproducible for nominal catalogs", I only have to set the random seed once per visit/instance catalog (because I'm calling this for all objects in the instance catalog) and then assume that 'nominal catalogs' came back in the same objid order (I know that's not really the same  .. unless there's a sort by objid somewhere, but it's probably close enough for me for now). So I guess I'll just set it when I call this, using the obshistid maybe.

python/lsst/sims/catUtils/mixins/PhotometryMixin.py
```diff
@@ -148,6 +148,35 @@ def get_lsst_photometric_uncertainties(self):
...
+        @ param [in] sigma is the FWHM of the distribution
+
+        @ param [out] visibility (None/1).
+        """
+        if len(magFilter) == 0:
+            return numpy.array([])
+        # Calculate the completeness at the magnitude of each object.
+        completeness = 1.0 / (1 + numpy.exp((magFilter - self.obs_metadata.m5[self.obs_metadata.bandpass])/sigma))
+        if randomSeed is not None:
+            numpy.random.seed(randomSeed)
```
danielsf: Note: if your query to fatboy is large enough, you could get your results in several chunks, in which case this method will be called once for each chunk, and the seed will keep getting reset.

You can have a member variable that exists just to check whether or not numpy.random has been seeded.  Something like:

if not hasattr(self, '_ssm_has_been_seeded'):
    numpy.random.seed(obs_metadata.mjd)
    self._ssm_has_been_seeded = True

Then the sequence will only be seeded once.

python/lsst/sims/catUtils/mixins/PhotometryMixin.py
```diff
@@ -148,6 +148,44 @@ def get_lsst_photometric_uncertainties(self):
...
+        if pre_generate_randoms:
```
danielsf: if you make a database query that is large enough to come back in multiple chunks, the randoms will be generated each time.

This should probably be

if pre_generate_randoms and not hasattr(self, 'randoms')

Also: maybe rename 'randoms' to 'ssm_randoms', just in case we end up adopting this solution for another use case

python/lsst/sims/catUtils/mixins/PhotometryMixin.py
```diff
@@ -148,6 +148,44 @@ def get_lsst_photometric_uncertainties(self):
...
+        @ param [in] pre_generate_randoms is an option (default False) to pre-generate a series of 12,000,000 random numbers
+           for use throughout the visibility calculation [the random numbers used are randoms[objId]].
+
+        @ param [out] visibility (None/1).
+        """
+        if len(magFilter) == 0:
+            return numpy.array([])
+        # Calculate the completeness at the magnitude of each object.
+        completeness = 1.0 / (1 + numpy.exp((magFilter - self.obs_metadata.m5[self.obs_metadata.bandpass])/sigma))
+        if randomSeed is not None:
```
danielsf: You should also add a member variable to track whether or not numpy.random has been seeded (so that it doesn't get seeded more than once in cases where the fatboy query comes back in chunks)

danielsf:
Sorry I did not think of this earlier:   should add a unit test that just tries to create one of these catalogs.  That way, if we ever change something (the way I am intending to change ObservationMetaData during this sprint) that breaks this API, the unit tests will catch that.

If you don't want to go through the trouble of creating dummy databases so that the unittest can run without a connection to fatboy, look at sims_catUtils/tests/testAllObjects.py for an example of how to write a unit test that connects to fatboy but ignores any errors that are raised because fatboy happens to be down (or you happen to be running on buildbot or Jenkins or any of our other CI machines that do not connect to fatboy).

If you would rather make this a separate issue, we can do that, too (and I will make sure to deal with that issue before changing ObservationMetaData's API)

rhiannonlynne:
I have a catalog generation script -- would that suffice to use as a unit
test (assuming I add the ignoring the errors?) It does wrap in a lot of the
mixins .. '

How are you changing observationmetadata -- just making all of the opsim
info available as a dictionary?

danielsf:
Yeah.  That is the kind of unit test I had in mind.  We just have to make sure that, if it fails for a real reason (rather than because we are not connected to fatboy), unittest complains.

In addition to adding the dict, I'm going to change ObservationMetaData so that the RA, Dec you give to the constructor is the actual pointing RA, Dec, instead of some abstract, unrefracted, above-the-atmosphere RA, Dec, as it is now.

PR #36, opened 2015-11-24
Bugfix/SIM-1595/CompoundCatalogDBObject

danielsf:


tests/testPhoSimCatalogs.py
```diff
@@ -79,12 +79,31 @@ def testCompoundCatalog(self):
...
-        testBulge = PhoSimCatalogSersic2D(self.bulgeDB, obs_metadata = self.obs_metadata)
-        testDisk = PhoSimCatalogSersic2D(self.diskDB, obs_metadata = self.obs_metadata)
-        testAgn = PhoSimCatalogZPoint(self.agnDB, obs_metadata = self.obs_metadata)
-        testStar = PhoSimCatalogPoint(self.starDB, obs_metadata = self.obs_metadata)
 
-        compoundCatalog = CompoundInstanceCatalog([testBulge, testDisk, testAgn, testStar],
+        # because the CompoundCatalogDBObject requires that database
+        # connection parameters be set in the input CatalogDBObject
+        # daughter class definitions, we have to declare dummy
+        # instances of our desired CatalogDBObject daughter classes
```
jbkalmbach: Aren't we declaring dummy classes here not instances? We shouldn't be sending instances into compoundCatalogDBObject, just daughter classes that will be instantiated from within compoundCatalogDBObject, right?

jbkalmbach:
Approved. The code looks good. Just a comment about the documentation.

tests/testPhoSimCatalogs.py
danielsf: good catch.  Will fix that

PR #38, opened 2015-12-17
Sniacatsim rebase

rbiswas4:
Pull request for SNIa catalog being added to sims_catUtils as mixin.

There are three key components:

snObject.py : contains the Object that describes SN properties in terms of a SALT model. This object contains multiple equivalent ways of completely specifying the properties of a SN, and computes the predicted flux and magnitudes of the SN as a function of time in different bandpasses.

snUniversalRules.py  Contains the population statistics of SNIa model parameters in the universe. Currently, fairly simple distribution with parameters are used, and the parameters of these distributions are user modifiable.  A default description has very simplistic choices of these parameters, creating a default description that is tied to nature through observations of such distributions is not currently implemented. 
 

sncat.py : Inherits from InstanceCatalog Objects, and meant to produce an instance catalog of supernovae, by using the population statistics on snUniversalRules to populate the catalog with instances of `snObject.SNObject`.

The examples directory contains examples of both snObject and snCatalogs as well as testSN.py which has unit tests for these modules.


python/lsst/sims/catUtils/mixins/snUniversalRules.py
```diff
@@ -0,0 +1,171 @@
...
+        '''
+        if not hasattr(self, '_midSurveyTime'):
+            midSurveyTime_default = 570000.0
+            self._midSurveyTime = midSurveyTime_default
+        return self._midSurveyTime
+
+    @midSurveyTime.setter
+    def midSurveyTime(self, mymidSurveyTime):
+        '''
+        set the value of suppressDimSN of the catalog
```
danielsf: Does this docstring belong with a different setter?  You are defining a setter for midSurveyTime, but the docstring says you are setting suppressDimSN.

python/lsst/sims/catUtils/mixins/snUniversalRules.py
```diff
@@ -0,0 +1,171 @@
...
+    badvalues
+    midSurveyTime
+    suppressDimSN
+    mjdobs
+    maxTimeSNVisible
+
+    """
+
+    @property
+    def averageRate(self):
```
danielsf: Is this a rate (with units of SNe per time) or an average time between SNe?

python/lsst/sims/catUtils/mixins/snUniversalRules.py
```diff
@@ -0,0 +1,171 @@
+import numpy as np
+
+__all__ = ['SNUniverse']
+
+class SNUniverse(object):
+    """
+    Mixin Class providing methods for distributing SN model parameters
+    conditioned  on the parameters of the hostgalaxy from catsim Catalogs
+
+    The base class must have the following attributes:
```
danielsf: Include a brief description of each of these required attributes.

python/lsst/sims/catUtils/mixins/snObject.py
```diff
@@ -0,0 +1,718 @@
...
+        # ie. no extinction, but this is not part of the API, so should not
+        # depend on it, set explicitly in order to unextincted SED from
+        # SNCosmo. We will use catsim extinction from `lsst.sims.photUtils`.
+
+        self.ModelSource = source
+        self.set(mwebv=0.)
+
+        # ra and dec if passed are assumed to be in degrees and converted into
+        # radians.
+        self._ra = ra
```
danielsf: You aren't converting into radians

python/lsst/sims/catUtils/mixins/snObject.py
```diff
@@ -0,0 +1,718 @@
...
+        # New Attributes
+        # statedict['lsstmwebv'] = self.lsstmwebv
+        statedict['_ra'] = self._ra
+        statedict['_dec'] = self._dec
+        statedict['MWE(B-V)'] = self.ebvofMW
+
+        return statedict
+
+    @staticmethod
+    def equivsncosmoParamDict(SNstate, SNCosmoModel):
```
danielsf: Can this be done without forcing the user to pass in an SNCosmoModel?  i.e., can we just instantiate a dummy SNCosmo model in this method (or will that not have the correct param_names)?

python/lsst/sims/catUtils/mixins/snObject.py
```diff
@@ -0,0 +1,718 @@
...
+        """
+        sncosmoParams = dict()
+        for param in SNstate.keys():
+            if param in SNCosmoModel.param_names:
+                sncosmoParams[param] = SNstate[param]
+        sncosmoParams['mwebv'] = SNstate['MWE(B-V)']
+        return sncosmoParams
+
+    @staticmethod
+    def sncosmoParamDict(SNstate, SNCosmoModel):
```
danielsf: Why is it necessary that this be a separate method?  The only difference between this and equivsncosmoParamDict is that this method does not set sncosmoParams['mwebv'].  Is that important?

python/lsst/sims/catUtils/mixins/snObject.py
```diff
@@ -0,0 +1,718 @@
...
+        -------
+        None
+
+        Examples
+        --------
+        >>> t = SNObject()
+        >>> t.set_MWebv(0.)
+        >>> 0.
+
+        .. note:: For a large set of SN, one may use fast `np.ndarray` valued
```
danielsf: This sentence makes it sound to me like you can pass an np.ndarray in for the value arg.  Is that true?  I though SNObject could only characterize one super nova at a time.

python/lsst/sims/catUtils/mixins/snObject.py
```diff
@@ -0,0 +1,718 @@
...
+        .. note: If there is an unphysical value of sed in
+        the wavelength range, it produces a flux of  `np.nan`
+        """
+
+        if magnitude is None:
+            mag = self.catsimBandMags(time=time, bandpassobject=bandpassobject)
+        else:
+            mag = magnitude
+
+        # mag = np.asarray([[mag]])
```
danielsf: remove unused code

python/lsst/sims/catUtils/mixins/snObject.py
```diff
@@ -0,0 +1,718 @@
...
+        bandpass = bandpassobject
+
+        if photParams is None:
+            photParams = PhotometricParameters()
+
+        # m5 = np.asarray([[m5]])
```
danielsf: remove unused code

python/lsst/sims/catUtils/mixins/snObject.py
```diff
@@ -0,0 +1,718 @@
...
+        Returns
+        -------
+        float
+
+        Examples
+        --------
+        .. note: If there is an unphysical value of sed in
+        the wavelength range, it produces a flux of  `np.nan`
+        """
+        # SEDfromSNcosmo = self.SNObjectSED(time=time,
```
danielsf: remove unused code

python/lsst/sims/catUtils/mixins/snObject.py
```diff
@@ -0,0 +1,718 @@
...
+        #                                 bandpass=bandpassobject)
+
+        if magnitude is None:
+            mag = self.catsimBandMags(time=time, bandpassobject=bandpassobject)
+        else:
+            mag = magnitude
+        flux = self.catsimBandFluxes(time=time, bandpassobject=bandpassobject)
+
+        # mag = np.asarray([[mag]])
```
danielsf: remove unused code

python/lsst/sims/catUtils/mixins/snObject.py
```diff
@@ -0,0 +1,718 @@
...
+        bandpass = bandpassobject
+
+        if photParams is None:
+            photParams = PhotometricParameters()
+
+        # m5 = np.asarray([[m5]])
```
danielsf: remove unused code

python/lsst/sims/catUtils/mixins/snObject.py
```diff
@@ -0,0 +1,718 @@
...
+        the wavelength range, it produces a flux of  `np.nan`
+        """
+        # SEDfromSNcosmo = self.SNObjectSED(time=time,
+        #                                 bandpass=bandpassobject)
+
+        if magnitude is None:
+            mag = self.catsimBandMags(time=time, bandpassobject=bandpassobject)
+        else:
+            mag = magnitude
+        flux = self.catsimBandFluxes(time=time, bandpassobject=bandpassobject)
```
danielsf: calling catsimBandMags and catsimBandFluxes means you are integrating the SED over the Bandpass twice.  Can you just do one and use the known relationship between flux and magnitude to get the other?

python/lsst/sims/catUtils/mixins/snObject.py
```diff
@@ -0,0 +1,718 @@
...
+        Examples
+        --------
+        .. note: If there is an unphysical value of sed in
+        the wavelength range, it produces a flux of  `np.nan`
+        """
+        SEDfromSNcosmo = self.SNObjectSED(time=time,
+                                          bandpass=bandpassDict['u'])
+        wavelen_step = np.diff(SEDfromSNcosmo.wavelen)[0]
+        SEDfromSNcosmo.flambdaTofnu()
+        f = SEDfromSNcosmo.manyFluxCalc(bandpassDict.phiArray,
```
danielsf: bandpassDict has its own equivalent of manyFluxCalc (BandpassDict.fluxListForSed).  Why aren't you using that?  Using the BandpassDict method will probably make this code easier to maintain after our sims_photUtils overhaul, as the BandpassDict API is much more likely to remain stable than is the Sed.manyFluxCalc API.

python/lsst/sims/catUtils/mixins/snObject.py
```diff
@@ -0,0 +1,718 @@
...
+        if self.ebvofMW is None:
+            raise ValueError('ebvofMW attribute cannot be None Type and must'
+                             ' be set by hand using set_MWebv before this'
+                             'stage, or by using setcoords followed by'
+                             'mwEBVfromMaps\n')
+
+        SEDfromSNcosmo.addCCMDust(a_x=ax, b_x=bx, ebv=self.ebvofMW)
+        return SEDfromSNcosmo
+
+    def catsimBandFluxes(self, time, bandpassobject):
```
danielsf: should this method be named 'catsimBandFlux', since it only calculates flux in one bandpass?

python/lsst/sims/catUtils/mixins/snObject.py
```diff
@@ -0,0 +1,718 @@
...
+        .. note: If there is an unphysical value of sed in
+        the wavelength range, it produces a flux of  `np.nan`
+        """
+        f = self.catsimManyBandFluxes(time,
+                                      bandpassDict,
+                                      observedBandPassInd)
+
+        return -2.5 * np.log10(f)
+
+    def catsimBandMags(self, bandpassobject, time):
```
danielsf: should this method be called 'catsimBandMag' since it only calculates the magnitude in one bandpass?

python/lsst/sims/catUtils/mixins/snObject.py
```diff
@@ -0,0 +1,718 @@
...
+        -------
+            float
+        Examples
+        --------
+        """
+        fluxRatio = self.catsimBandFluxes(bandpassobject=bandpassobject,
+                                          time=time)
+        return -2.5 * np.log10(fluxRatio)
+
+    def catsimADU(self, time, bandpassDict,
```
danielsf: I think you should keep the same convention you established for magnitudes and fluxes:

catsimBandADU calculates the ADU in one bandpass

catsimManyBandADU calculates the ADU in all the bandpasses of a BandpassDict

python/lsst/sims/catUtils/mixins/snObject.py
```diff
@@ -0,0 +1,718 @@
...
+                  photParams=None,
+                  observedBandPassInds=None):
+        """
```
danielsf: This docstring should be made consistent with the other docstrings (in terms of formatting)

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -0,0 +1,302 @@
...
+        '''
+        return np.nan
+
+    @property
+    def suppressDimSN(self):
+        '''
+        Boolean to decide whether to output observations of SN that are too dim
+        should be represented in the catalog or not. By default set to True
+        '''
+        # self._suppressDimSN = True
```
danielsf: remove unused code

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -0,0 +1,302 @@
...
+    def suppressDimSN(self, suppressDimSN):
+        '''
+        set the value of suppressDimSN of the catalog
+
+        Parameters
+        ----------
+        value : Boolean, mandatory
+            Value to set suppressDimSN to
+        '''
+        # if suppressDimSN is None:
```
danielsf: remove unused code

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -0,0 +1,302 @@
...
+    def suppressHighzSN(self):
+        '''
+        Boolean to decide whether to output information corresponding to SN
+        at redshift above self.maxz
+        '''
+
+        return True
+
+    @property
+    def maxTimeSNVisible(self):
```
danielsf: Should this be related to the _averageRate property in SNUniverse?

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -0,0 +1,302 @@
...
+        '''
+        The catalog will provide values for SN flux (even if zero according to
+        model) for peak \pm maxTimeVisibilityforSN
+        '''
+        return 100.
+
+    @property
+    def maxz(self):
```
danielsf: Should this be a settable property so that users can change it if they want to?

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -0,0 +1,302 @@
...
+            Value to set suppressDimSN to
+        '''
+        # if suppressDimSN is None:
+        #    self._suppressDimSN = True
+        # else:
+        self._suppressDimSN = suppressDimSN
+        return self._suppressDimSN
+
+    @property
+    def suppressHighzSN(self):
```
danielsf: Should users be able to turn this on and off?

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -0,0 +1,302 @@
...
+    def lsstBandpassDict(self):
+        return BandpassDict.loadTotalBandpassesFromFiles()
+
+    @astropy.utils.lazyproperty
+    def observedIndices(self):
+        bandPassNames = self.obs_metadata.bandpass
+        return [self.lsstBandpassDict.keys().index(x) for x in bandPassNames]
+
+    @astropy.utils.lazyproperty
+    def lsstpbase(self):
```
danielsf: It does not look like you use this property.  If so, I think it should be dropped.

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -0,0 +1,302 @@
...
+        `np.ndarray` of coordinara, dec, z, vra, vdec, and vr
+
+        '''
+        hostra, hostdec, hostz = self.column_by_name('raJ2000'),\
+            self.column_by_name('decJ2000'),\
+            self.column_by_name('redshift')
+        snra, sndec, snz, snvra, snvdec, snvr = self.SNCoordinatesFromHost(
+            hostra, hostdec, hostz)
+
+        return ([snra, sndec, snz, snvra, sndec, snvr])
```
danielsf: sndec appears twice in the reutrn np.array.  Second time is supposed to be snvdec.

python/lsst/sims/catUtils/mixins/snUniversalRules.py
```diff
@@ -0,0 +1,171 @@
+import numpy as np
```
danielsf: The methods in SNUniverse need more complete docstrings (for example, the input parameters for SNparamDistFromHost are not defined)

python/lsst/sims/catUtils/mixins/snUniversalRules.py
```diff
@@ -0,0 +1,171 @@
...
+        snvra = np.zeros(numhosts)
+        snvdec = np.zeros(numhosts)
+        snvr = np.zeros(numhosts)
+
+        if self.suppressHighzSN:
+            snz = np.where(snz > self.maxz, np.nan, snz)
+    
+        return snra, sndec, snz , snvra, snvdec, snvr 
+
+    def SNparamDistfromHost(self, hostz, hostid, hostmu):
```
danielsf: Camel case is inconsistent.  You should capitalize 'From' in the method name.

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -0,0 +1,302 @@
...
+            self.column_by_name('decJ2000')
+
+        SNobject = SNObject()
+
+        sedlist = []
+        for i in range(self.numobjs):
+            SNobject.set(z=_z[i], c=c[i], x1=x1[i], t0=t0[i], x0=x0[i])
+            SNobject.setCoords(ra=ra[i], dec=dec[i])
+            SNobject.mwEBVfromMaps()
+            sed = SNobject.SNObjectSED(time=self.obs_metadata.mjd,
```
danielsf: did you mean to use self.mjdobs as the time kwarg?

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -0,0 +1,302 @@
...
+    def get_snparams(self):
+        hostz, hostid, hostmu = self.column_by_name('redshift'),\
+            self.column_by_name('snid'),\
+            self.column_by_name('cosmologicalDistanceModulus')
+
+        vals = self.SNparamDistfromHost(hostz, hostid, hostmu)
+
+        return (vals[:, 0], vals[:, 1], vals[:, 2], vals[:, 3])
+
+    def get_SNsed(self):
```
danielsf: defining this as a get_ method means that it is possible for a user to try to write it out into a catalog.  I do not know what will happen if they try to write SEDs out to an InstanceCatalog.  Can you change the name of this method to something like 'loadSEDs' so that there is no risk of SNsed being added to column_outputs?

python/lsst/sims/catUtils/mixins/snUniversalRules.py
rbiswas4: You are correct. This docstring is wrong and needs to be fixed. Fixed in  51cf55d289ec0adf8e869af53eb5ac03b4450082

python/lsst/sims/catUtils/mixins/snObject.py
rbiswas4:
```I think you can get away with this because setCoords sets self._ra and self._dec to radians. Maybe you should just get rid of this assignment of self._ra and self._dec.
```
Yes, that was the plan. The reason I had something before that was to check if ra, dec were provided or were NoneTypes (default). This check could be done on the input parameters  ra,dec. 
But I wanted to provide values in SNstate too (which right now is taken directly from self._ra, self._dec.

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -0,0 +1,302 @@
...
+        c, x1, x0, t0, _z, ra, dec = self.column_by_name('c'),\
+            self.column_by_name('x1'),\
+            self.column_by_name('x0'),\
+            self.column_by_name('t0'),\
+            self.column_by_name('redshift'),\
+            self.column_by_name('raJ2000'),\
+            self.column_by_name('decJ2000')
+
+        SNobject = SNObject()
+        bandname = self.obs_metadata.bandpass
```
danielsf: since it is possible that obs_metadata.bandpass is a list (at least, for now it is), you should probably have a check here that raises an error if obs_metadata.bandpass is a list (since this getter appears to be only able to return the flux and magnitude in a single bandpass)

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -0,0 +1,302 @@
...
+        # Initialize return array
+        vals = np.zeros(shape=(self.numobjs, 4))
+
+        for i, v in enumerate(vals):
+            SNobject.set(z=_z[i], c=c[i], x1=x1[i], t0=t0[i], x0=x0[i])
+            SNobject.setCoords(ra=ra[i], dec=dec[i])
+            SNobject.mwEBVfromMaps()
+
+            # Calculate fluxes
+            flux = SNobject.catsimBandFluxes(time=self.mjdobs,
```
danielsf: calling catsimBandFluxes, catsimBandMags, catsimBandFluxError, and catsimBandMagError involves calculating a bunch of things twice.  Is there a way this can be done with caching so that the calculation is more efficient.

At the very least: maybe you could use self._actually_calculated_columns to determine if the user is actually asking for a column and refrain from calculating anything the user doesn't really want.

danielsf: _actually_calculated_columns is a member variable of InstanceCatalog that contains the names of columns that the user explicitly asked for and the columns that those explicitly requested columns depend on.

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -0,0 +1,302 @@
...
+            vals[i, 2] = flux_err
+            vals[i, 3] = mag_err
+        return (vals[:, 0], vals[:, 1], vals[:, 2], vals[:, 3])
+
+    @compound('flux_u', 'flux_g', 'flux_r', 'flux_i', 'flux_z', 'flux_y',
+              'mag_u', 'mag_g', 'mag_r', 'mag_i', 'mag_z', 'mag_y',
+              'adu_u', 'adu_g', 'adu_r', 'adu_i', 'adu_z', 'adu_y', 'mwebv')
+    def get_snfluxes(self):
+
+        c, x1, x0, t0, _z, ra, dec = self.column_by_name('c'),\
```
danielsf: can you have get_snbrightness depend on get_snfluxes (so that code is not duplicated)?

python/lsst/sims/catUtils/mixins/snObject.py
danielsf: Yeah.  It's probably safer to just do the test on the input parameters.  Otherwise, it looks like you forgot to call np.radians()

danielsf:
I've looked at all of the scripts in mixins/

I will look at the example iPython notebooks tomorrow.

python/lsst/sims/catUtils/mixins/snObject.py
rbiswas4: And initialize self._ra = None ?

python/lsst/sims/catUtils/mixins/snObject.py
rbiswas4: It can be done, but I don't think it is a good idea. I don't think the user will be using this function as part of the class a lot, but might use it as a static method without instantiating the class more often. Thus, allowing the user to set the dummy is more useful.

python/lsst/sims/catUtils/mixins/snUniversalRules.py
rbiswas4: You are right. That is a badly named variable. Can we call it SNfrequency and invert it ?

python/lsst/sims/catUtils/mixins/snObject.py
rbiswas4: I can make this method 'private'.

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: No. maxTimeVisible is a couple of months during which the SN is bright. Not using would mean that we would get 0 flux from the SN in files for very long times. This is correct in principle but will only be a pain to deal with. The other attribute (which as we discussed we should change variable names for) is how frequently a SN goes off in a galaxy. This is roughly once a century. 

for the record: There is a bit of a cheat going on here: The SN models we have do not go on for ever and have values in the range on -30, 50 days in the rest frame. Beyond that the flux is non-zero but really small and does not matter in practice. I have not done the study of looking at the correct extrapolation of the model beyond this range.

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: It is. You can always do 
```
s = sncat(...)
s.maxz = 1.8
```
 At some point I want to make maxz a function later. Of course, we could argue that we could simply have introduced an attribute rather than a property, and change that when we actually decide to make it a function.

python/lsst/sims/catUtils/mixins/snUniversalRules.py
rbiswas4: Done. 0d70a3fa68ea80c128115c75470d0a78ae7aae51

python/lsst/sims/catUtils/mixins/snObject.py
rbiswas4: Agreed. This seems to be wrong due to the reason you mention. I cannot remember any rationale for having that line. I am getting rid of it in 4b6b9814ea98e86f89517bd575794134388fcb24

python/lsst/sims/catUtils/mixins/snObject.py
rbiswas4: Done: 5b8647a9525d941562e559ae4ac74254eb68cd36

python/lsst/sims/catUtils/mixins/snObject.py
rbiswas4: Done in commit

python/lsst/sims/catUtils/mixins/snObject.py
rbiswas4: Done in commit

python/lsst/sims/catUtils/mixins/snObject.py
rbiswas4: Done in commit

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: Done in commit

python/lsst/sims/catUtils/mixins/snObject.py
rbiswas4: Done in commit

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: Done in commit

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: Similar logic. But here, I agree that we could make this an attribute rather than use property

python/lsst/sims/catUtils/mixins/snUniversalRules.py
danielsf: Sure.  Or SNperiod.  Whatever is most convenient for you (since you're the one using it).

python/lsst/sims/catUtils/mixins/snObject.py
danielsf: I guess so.

python/lsst/sims/catUtils/mixins/snObject.py
danielsf: What use case do you imagine for this method?  It looks to me like this is trying to produce a ```__repr___``` for the SNObject.  Should we just define ```__repr___``` and use that?

python/lsst/sims/catUtils/mixins/snObject.py
danielsf: Again: I guess I'm not sure what these methods are for.

python/lsst/sims/catUtils/mixins/sncat.py
danielsf: Okay.  I did forget the distinction between supernova rate and the duration of a specific supernova.

python/lsst/sims/catUtils/mixins/sncat.py
danielsf: I think you need to define a setter for maxz in order for the code you just described to work.

python/lsst/sims/catUtils/mixins/sncat.py
danielsf: Again: I think what you intended will only work if you define a setter for suppressHighzSN

danielsf:
I also think snObject.py and snUniversalRules.py should be moved into their own sub directory of sims_catUtils (something like lsst/sims/catUtils/supernovae/) simply because they are not mixins, but classes that support the SNIaCatalog mixin.

python/lsst/sims/catUtils/mixins/snObject.py
rbiswas4: The closest thing to a __repr__ here is the SNstate attribute. I might want to think of doing that at a later time (make a JIRA issue out of it now) or take the plunge.  But this function is not for that. 

The use for this function is to set values in a SNCosmo model from our SNObject instances. This allows people to get parameter dicts that work with the rest of SNCosmo for related models. This is a useful function to have once you move out of catalog only space.

python/lsst/sims/catUtils/mixins/snUniversalRules.py
rbiswas4: I think SNperiod will sound like ~ 50 days to most people. So I will stick to SNfrequency.

danielsf:
Is it necessary that the .png files in examples/SNCatalog_example/ be a part of the repository?  I worry that the repo history could explode if those figures get updated with any frequency.

Similarly: can we run  ```Cell >> All output >> Clear``` on the iPython notebooks before committing them to the repo?

danielsf:
The API for ObservationMetaData has changed since you wrote these iPython notebooks.  Most notably, unrefractedRA and unrefractedDec have become pointingRA and pointingDec.

danielsf:
There seem to be a lot of methods (```samplePatchOnSphere```, ```sample_obsmetadata```, ```cleanDb```) that are duplicated in your iPython notebooks.  Do you want to define them in ```sims_utils``` so that you can just import them?

Also: look at the ```sampleSphere``` method in ```sims_catalogs_generation/python/.../utils/testUtils.py``` That might be what you want from your ```samplePatchOnSphere``` method.

rbiswas4:
> Is it necessary that the .png files in examples/SNCatalog_example/ be a part of the repository? I worry that the repo history could explode if those figures get updated with any frequency.

Similarly: can we run Cell >> All output >> Clear on the iPython notebooks before committing them to the repo?

tangent: I wonder if github has a better way to reply to issues that do not pertain to lines (or maybe they should just be issues?)

Now for your question: I think this is something to decide for all LSST sims/DM examples. And the answer is not clear to me, because of the problems you are bringing up.

- Not having the pngs/ outputs has the disadvantage that people will not be able to go to github and render the page as a tutorial. You might say that they could run it, but there is value in being able to see what it was meant to be even before running them.
- I have seen people using markdowns instead, (and I tried this too), but I don't see how to keep them uptodate in an easy way.

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: I agree that this would be a desirable change. Since it will not result in a change in User API, can we make this a JIRA issue for now, and figure out how to do this best later?

python/lsst/sims/catUtils/mixins/snObject.py
rbiswas4: Done: 5ecc0860ac96d539ded796b946ad22b59b14d882

rbiswas4:
> The API for ObservationMetaData has changed since you wrote these iPython notebooks. Most notably, unrefractedRA and unrefractedDec have become pointingRA and pointingDec.

Fixed to match new API in /examples/SNCatalog_example.ipynb. This is the only place where it is necessary? : 458226a19d3fc0ffce4c27d8fe48fbc09a531876

rbiswas4:
> There seem to be a lot of methods (samplePatchOnSphere, sample_obsmetadata, cleanDb) that are duplicated in your iPython notebooks. Do you want to define them in sims_utils so that you can just import them?

Yes. I think unifying these is what we want. We should discuss if utils already has something of this kind.
(like your next suggestion).
Also: look at the sampleSphere method in sims_catalogs_generation/python/.../utils/testUtils.py That might be what you want from your samplePatchOnSphere method.

python/lsst/sims/catUtils/mixins/snObject.py
danielsf: I see.

My original point was that SNCosmoModel appears to only be used to determine what parameters belong in the output dict.  I am proposing that we not make SNCosmoModel an argument, and instead have the function work like

```
def equivsnCosmoParamDist(SNstate):
    sncosmoParams = dict()
    dummy = sncosmo.model()
    for param in SNstate.keys():
        if param in dummy.param_names:
            sncosmoParams[param] = SNstate[param]
            sncosmoParams['mwebv'] = SNstate['MWE(B-V)']
     return sncosmoParams
```

python/lsst/sims/catUtils/mixins/sncat.py
danielsf: Sure.

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -0,0 +1,298 @@
...
+
+    override_formats = {'snra': '%8e', 'sndec': '%8e', 'c': '%8e',
+                        'x0': '%8e'}
+    for var in variables:
+        override_formats[var] = '%8e'
+
+    cannot_be_null = ['x0', 'z', 't0']
+
+    @astropy.utils.lazyproperty
+    def mjdobs(self):
```
danielsf: Be aware: I just merged a change to sims_utils.  ObservationMetaData.mjd is now an instance of the class ModifiedJulianDate, which is a thin wrapper of astropy time.  You probably want to return something like self.obs_metadata.mjd.TAI here.

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: Done: dfbb0ec3965c06210a2bbe034c280c8a855d51d7

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: Done: dfbb0ec3965c06210a2bbe034c280c8a855d51d7

python/lsst/sims/catUtils/mixins/snUniversalRules.py
rbiswas4: Done : 66f49b72ccaa70deca225431fe8f09dc523a8846

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: 6eef5061f4c3a2cebb9cd1282084a3aa3e97ca11

python/lsst/sims/catUtils/mixins/snUniversalRules.py
rbiswas4: Done : 6eef5061f4c3a2cebb9cd1282084a3aa3e97ca11

python/lsst/sims/catUtils/mixins/snObject.py
rbiswas4: Done: 53c8eb3914ce966889c23c078d8f219fa365eadf

python/lsst/sims/catUtils/mixins/snUniversalRules.py
rbiswas4: Done : a8113be0087cad77c1f62ba862db6602098ea3e2

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: Done: 38488537030f88e3eb345bd0b52020813498063d

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: Done: b194dd2c13e200c3001967511688e7ac328e3ad

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: Done: 0b982f7e0ce81896c18b1bb5a5005eeab9cf8ba4

python/lsst/sims/catUtils/mixins/snUniversalRules.py
rbiswas4: defined these. Though the doc strings could still be improved 

7fe7aee65af36b820a08da6ba2511fd9f7a84af

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: Done: 6873a6007c8dba759598e623a8b09ffec1da2fd1

python/lsst/sims/catUtils/mixins/snObject.py
rbiswas4: Done: 
353a122e92dda84c43197bc886a4256dedaffe61

python/lsst/sims/catUtils/mixins/snObject.py
rbiswas4: Done : 353a122e92dda84c43197bc886a4256dedaffe61

python/lsst/sims/catUtils/mixins/snObject.py
rbiswas4: This was done earlier ...can't remember commit

python/lsst/sims/catUtils/mixins/snObject.py
rbiswas4: catsimManyBandADU or catsimManyBandADUs ?

rbiswas4: Done : to ADUs 967024aabe2d08acc7556eec950e8386da281829

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: Done.

python/lsst/sims/catUtils/mixins/snObject.py
rbiswas4: Done: bd5294b9fa0dc7fdfdc5c9da3b8859cba23ed7b2

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: Done: bd5294b9fa0dc7fdfdc5c9da3b8859cba23ed7b2

rbiswas4:
> I also think snObject.py and snUniversalRules.py should be moved into their own sub directory of sims_catUtils (something like lsst/sims/catUtils/supernovae/) simply because they are not mixins, but classes that support the SNIaCatalog mixin.

Done : 79a49d36b166b8e8ee9596b1d54615d1f4b58221

python/lsst/sims/catUtils/mixins/snObject.py
rbiswas4: Will be done in a different pull request

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: Will be done in a different pull request

rbiswas4:
> There seem to be a lot of methods (samplePatchOnSphere, sample_obsmetadata, cleanDb) that are duplicated in your iPython notebooks. Do you want to define them in sims_utils so that you can just import them?

samplePatchOnSphere, sample_obsmetadata moved to sims.utils

rbiswas4:
There seem to be a lot of methods (samplePatchOnSphere, sample_obsmetadata, cleanDb) that are duplicated in your iPython notebooks. Do you want to define them in sims_utils so that you can just import them?
Yes. I think unifying these is what we want. We should discuss if utils already has something of this kind.
(like your next suggestion).
Also: look at the sampleSphere method in sims_catalogs_generation/python/.../utils/testUtils.py That might be what you want from your samplePatchOnSphere method.


I added the sampleSphere and sample_obsmetadata methods to sims_utils with a similar branch name

python/lsst/sims/catUtils/mixins/snObject.py
rbiswas4: Now tracked on Jira : https://jira.lsstcorp.org/browse/SIM-1723

rbiswas4:
@danielsf I think all of the crucial issues are addressed. A few improvement issues are left as JIRA issues with links so that they can be tracked. There is also an associated pull request in sims_utils.

danielsf:
In SNObject_example.ipynb I get

```ImportError: No module named seaborn```

in the first cell

danielsf:
also in SNObject_example.ipynb, ```from utils import plotutils as pu``` gives

```ImportError: No module named utils```

rbiswas4:
Sorry about that.

Ok. I am getting rid of the seaborn import. Not really important.

Can I add the plotUtils to sims_utils or do you want me to split up plots?
(This is  just a wrapper to matplotlib I use to set up Output 33, which
shows the trend and a differential slice)

danielsf:
In GalaxyCatalog.ipynb, make include some documentation at the top making sure that users know they need to have enigma_1189_sqlite.db on the correct path for the notebook to find it.

danielsf:
Let's not put plotting utilities in sims_utils just yet.  That seems too high level for that package.  Leave it in sims_catUtils.

danielsf:
good to merge

PR #40, opened 2016-01-06
No bug to fix after all ... 

rhiannonlynne:
it turned out, I was confused about what units were being used where, but that everything was ok. 
I added more comments regarding what units things are in, and added a test of the actual value of the trailing losses to the unit test. 

This could be merged or not .. it was useful for me to realize that things were ok.


danielsf:
go ahead and merge

PR #41, opened 2016-01-07
Bugfix/sim 1664/pointing

danielsf:


python/lsst/sims/catUtils/mixins/AstrometryMixin.py
```diff
@@ -92,6 +94,11 @@ class AstrometryGalaxies(AstrometryBase):
     This mixin contains a getter for the corrected RA and dec which ignores parallax and proper motion
```
jbkalmbach: This is confusing to me. Doesn't anything that calls _observedFromICRS like the 2nd and 3rd getter in this mixin take these effects into account?

python/lsst/sims/catUtils/mixins/AstrometryMixin.py
```diff
@@ -147,12 +154,29 @@ def get_phoSimCoordinates(self):
     def get_observedCoordinates(self):
         return self.observedStellarCoordinates()
 
+    @compound('raICRS', 'decICRS')
```
jbkalmbach: I think the docstring for this mixin should be updated as well since it seems that originally it only contained one getter but obviously contains multiple now.

jbkalmbach:
Ok

python/lsst/sims/catUtils/mixins/AstrometryMixin.py
danielsf: Yes, but because they are not specified as kwargs, parallax and proper motion are set to None.

Basically, I'm trying to say that this getter is for objects that do not have a parallax or proper motion defined.

I will try to rewrite it.

python/lsst/sims/catUtils/mixins/AstrometryMixin.py
danielsf: Done

PR #42, opened 2016-01-13
Feature/sim 1054/new sed library

danielsf:
This pull request adds a unit test to make sure that users are running with the expected version of sims_sed_library installed.


tests/testFileMap_existence.py
```diff
@@ -0,0 +1,133 @@
+"""
```
yalsayyad: I thought about this a little more. The majority (all without the commented block) of the asserts here are actually testing the functionality of `sims_utils` not `sims_catUtils`'   I know we need to make sure that users have the new sed library installed before they go accessing the LSSTCATSIM stars.  But unit tests aren't REALLY the best way to check version dependencies. It seems to me like the cleanest way would be to add a version requirement to the .table file like `setupRequired(sims_sed_library >= 1.2.3)`  (though like you were saying currently the only explicit dependency on sims_sed_library is in photUtils, so you'd then have either require that catUtils depends on the version of photUtils that depends on the new sims_sed_library OR make catUtil's dependency on sims_sed_library explicit.  I vote for making the sims_catUtil's dependency on sims_sed_library explicit, since that was intention behind this unit test.

The other day you might have been telling me that we can't version sims_sed_library until we move it to lfs...  If that is the case, perhaps we should add a note here that this unit test is just a temporary way to check the version of the sed_library.

danielsf:
I agree that we should make the sims_catUtils dependency on sims_sed_library more explicit.  I imagined the work for SIM-1714 would involve going through all of the sims packages and making sure their dependencies were explicitly declared.

Regarding this unit test: I imagined it as a final fail-safe that would trigger if, for some reason, our version resolution methods failed to setup the correct sims_sed_library (otherwise, users won't realize something is wrong until they try to actually generate a catalog, and even then only if the catalog contains one of the new SEDs).  If you think I am being too paranoid, I can jettison this test.  I tend to err on the side of paranoia.

PR #43, opened 2016-01-26
Bugfix/sim 1732/mlt pho sim

danielsf:
When we updated the sims_sed_library last week, the MLT dwarves were matched to SED files each with 58,500 lines in them.  PhoSim only allows SED files to have 24,999 lines in them, so I added a sub-folder to sims_sed_library/starSED/ that contains the MLT dwarf spectra clipped to meet this requirement.  I then wrote a SpecFileMap sub-class specifically for the PhoSim InstanceCatalogs so that MLT dwarf spectra names get mapped to that sub-directory.

There are pull requests in

sims_utils
sims_catalogs_measures
sims_catUtils
sims_sed_library

supporting this pull request


yoachim:
looks good.

PR #44, opened 2016-01-27
kraken_1042 compatibility for ObservationMetaDataGenerator

danielsf:
The change in OpSim DB schema was just that 'finSeeing' was replaced with 'FWHMgeom' and 'FWHMeff'.  I have added code to the ```__init___``` of ````ObservationMetaDataGenerator``` to detect which column exists in the database and use that column to get seeing data for the resulting ```ObservationMetaData```s.


SimonKrughoff:
The summary of the commit message is too long.  [Here](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) is a good summary of commit message formatting.  I know it seems pedantic, but many tools treat the first line specially and if it is longer than 50 characters, it's hard to read.

danielsf:
I shortened the commit message.  It appears on one line in ```git log```

SimonKrughoff:
I have a .vimrc that will change the highlighting when my summary message is too long.  It's very useful.

PR #45, opened 2016-01-27
Feature/sim 1740/de precession

danielsf:
This makes the "de-precession" transformations the default for RA, Dec output to PhoSim InstanceCatalogs


SimonKrughoff:
These commit messages are excellent.  Thanks very much for taking to heart my earlier suggestions.  It goes a long way toward making clear your intent.

PR #48, opened 2016-02-01
Split SNIaCatalog into two classes

rbiswas4:
Now we have the following classes
- SNFunctionality : which calculates fluxes, magnitudes, errors etc for
  SNObjects that have a certain parameter set
- SNIaCatalog  which has the catalog of SN parameter sets.

This is useful in defining mulitple ways of creating the catalog of SN
parameter sets efficiently reusing code for calculation of quantities
from these parameters.

	modified:   ./examples/SNCatalog_example.ipynb (cosmetic, ipython changes)
	modified:   ./python/lsst/sims/catUtils/mixins/sncat.py

Ran both ./examples/testSN.py and ./SNCatalog_example.ipynb correctly. Checked that branch builds on Jenkins: https://ci.lsst.codes/job/stack-os-matrix/7852/


danielsf:
I think there need to be two mixins.

```SNparamsFromGalTiled``` which defines ```get_angularCoordinates()``` and ```get_snparams()```, and ```SNFunctionality``` which includes all of the other methods, but does not inherit from ```InstanceCatalog```.

You can then define a the class

```
class SNIaCatalog(InstanceCatalog, SNFunctionality, SNparamsFromGalTiled):
    pass
```

That way, someone who wants to use the Twinkles supernova catalog (and thus does not want ```get_angularCoordinates()``` or ```get_snparams()```) can define a class

```
class myTwinklesCatalog(InstanceCatalog, SNFunctionality):
    pass
```

without having to redefine ```suppressHighzSN```, ```maxtimeSNVisible``` and all of the other variables defined at the top of your definition for ```SNIaCatalog```.  Meanwhile, someone who wants to define the SN parameters in a different way can define

```
class myCustomSNCatalog(InstanceCatalog, SNFunctionality):

    def get_snparams(self):
        ...customised sn parameter code...
```

Basically, we want to make it as easy as possible for users to mix-and-match the super nova methods in their own ```InstanceCatalog``` classes.

rbiswas4:
Agree completely ... and have kind of gone down that path already.

danielsf:
looks good to merge

PR #49, opened 2016-02-08
The negative values in un-rectified SED of SALT2 (and SALT2-extended) at

rbiswas4:
extreme regions of the SNIa phase space can be a problem for codes that
sample the SED like phoSim. Therefore, this change implements:

rectified SED(lambda) = modelSED if SED(lambda) >= 0
                      = 0 if modelSED(lambda) < 0

This will cause the band flux and band mags (model values to be slightly
biased, but this should not matter much as they are truly small hen
compared to reasonable sky noise

	modified:   examples/testSN.py
	modified:   python/lsst/sims/catUtils/supernovae/snObject.py


examples/testSN.py
```diff
@@ -9,7 +9,9 @@
         implementations of extinction using OD94 model.)
     - Band Flux for extincted SED in r Band
     - Band Mag for extincted SED in r Band
-
+    - rectification of SED: while the SEDs from the model can go negative
+      resulting in negative model fluxes and adus. Test that if rectifySEDs in
```
danielsf: typo: "in on" should be "is on"

python/lsst/sims/catUtils/supernovae/snObject.py
```diff
@@ -454,7 +473,8 @@ def SNObjectSED(self, time, wavelen=None, bandpass=None,
 
         # Set SED to 0 beyond the model phase range, will change this if
         # SNCosmo includes a more sensible decay later.
-        if (time > self.mintime()) & (time < self.maxtime()):
+        if (time > self.mintime()) & (time < self.maxtime())\
+                & (self.modelOutSideRange=='zero'):
```
danielsf: I don't understand this if statement.  The first argument (time>self.mintime() & time<self.maxtime()) guarantees that the model is inside the allowable time range.  Why do you need to test on self.modelOutSideRange?

python/lsst/sims/catUtils/supernovae/snObject.py
```diff
@@ -121,9 +122,27 @@ def __init__(self, ra=None, dec=None, source='salt2-extended'):
         self.ebvofMW = None
         if self._hascoords:
             self.mwEBVfromMaps()
+
+
+        # Behavior of model outside range :
+        # if 'zero' then all fluxes outside are set to 0.
+        self._modelOutSideRange = 'zero'
```
danielsf: You should specify that, in this case, 'range' refers to a range in time, rather than a range in wavelength.

python/lsst/sims/catUtils/supernovae/snObject.py
danielsf: Wait.  I think I see what you are doing, but I don't think the logic will work the way you want it.  As it is, both conditionals must be true in order for flambda to be non-zero, so, if time is between mintime and maxtime, but self.modelOutSideRange is not 'zero' (i.e. the first conditional is True but the second conditional is False), flambda will be left as an array of zeros.  You want flambda to always be non-zero if time is between mintime and maxtime, regardless of what self.modelOutSideRange is, and flambda to be zero only if both time is outside of the valid range and self.modelOutSideRange is 'zero'.  Maybe something like:

```if (time<self.mintime() || time>self.maxtim()) and self.modelOutSideRange=='zero':
    flambda = np.zeros()
else:
    actually set flambda
```

examples/testSN.py
rbiswas4: Fixed

python/lsst/sims/catUtils/supernovae/snObject.py
rbiswas4: True, done

python/lsst/sims/catUtils/supernovae/snObject.py
rbiswas4: Did something similar.

rbiswas4:
OK ready for another look

danielsf:
commit 0498d9199 seems to have undone the changes in commit e639879

Why did you merge the branch into itself?

PR #51, opened 2016-02-14
Added a function to obtain the rest frame Sed source for phosim/galsim

rbiswas4:
This adds a function to obtain the rest frame SN spectrum, 
Within the model range, 
- It should match the source, except it takes on zero values if the source is negative.
- It goes to 0 outside the temporal range at which the model is defined
- It goes to np.nan outside the wavelength range at which the model is defined.


danielsf:
It looks like some of the commits in this pull request duplicate the commits in #49 .  How easy will it be to merge that, reimplement the actual changes from this branch on a new branch, and then issue a new pull request?

Sorry it took me so long to get to this.

rbiswas4:
This is superseded by #69

PR #54, opened 2016-05-20
Feature/sn object sed regions

rbiswas4:
minor BugFix, Speedups and additional feature of SED rectification in `sims_catUtils.supernovae.snobject.SNObject` + new tests to check feature of SED rectification.

BugFix : A bitwise and changed to and.
Speedup: When the time of observation is outside the temporal range of the model, and uses the modelOutSideTemporalRange 'zero' where the SEDs are 0, speed catsimBandflux calculations by returning 0 rather than assigning zeros to the SED, and doing a bandflux calculation.

SED rectification: At regions where the SALT model is poorly sampled in the training set, the SED can go negative. This forces such negative values to 0, which is important for tools like phosim that will sample the SED to generate photons. Tests are added.  Jenkins testbuild https://ci.lsst.codes/job/stack-os-matrix/11442/ passes. 

Note: This is a duplicate of a previous mangled pull request https://github.com/lsst/sims_catUtils/pull/49. That PR has a few more commits and I will delete the branch later.


danielsf:
Looking at the 'Console Output' for your Jenkins build

https://ci.lsst.codes/job/stack-os-matrix/11442/label=centos-7/console

 it looks like Jenkins built ```master``` rather than this branch

```
         meas_astrom: 2016_01.0-11-gc94ecbd+8 (already installed).
          pipe_tasks: 2016_01.0-46-gc25bc54+5 (already installed).
         obs_lsstSim: 2016_01.0-19-g85d3ac2+12 ok (28.2 sec).
            sims_maf: 2.1.3-39-g0014a03+13 ok (20.4 sec).
       sims_catUtils: 2.1.2-13-g1235fe2+1 .................ok (74.1 sec).
```

Are you sure you asked Jenkins to rebuild ```feature/SNObject_SEDregions```?

danielsf:
I think pandas has been incorporated into Jenkins now.  We should open a new ticket to try to move ```testSN.py``` out of ```examples/``` into ```tests/``` (or was there something other than pandas that was preventing that move?)

examples/testSN.py
```diff
@@ -11,23 +11,23 @@
...
-from lsst.sims.photUtils import Bandpass
+from lsst.sims.photUtils.PhotometricParameters import PhotometricParameters
+# from lsst.sims.photUtils import Bandpass
 from lsst.sims.photUtils import BandpassDict
 from lsst.sims.utils import ObservationMetaData
 from lsst.sims.utils import spatiallySample_obsmetadata as sample_obsmetadata
 from lsst.sims.catUtils.utils import ObservationMetaDataGenerator
 from lsst.sims.catalogs.generation.db import CatalogDBObject
-from lsst.sims.catalogs.measures.instance import InstanceCatalog
+# from lsst.sims.catalogs.measures.instance import InstanceCatalog
```
danielsf: If this is no longer needed, you should delete it.

examples/testSN.py
```diff
@@ -94,7 +94,62 @@ def test_SNstatenotEmpty(self):
...
+                                    bandpass=self.lsstBandPass['r'])
+            if any(sed.flambda < 0.):
+                badTimes.append(time)
+        # Check that there are negative SEDs
+        assert(len(badTimes) > 0)
+        snobj.rectifySED = True
+        for time in badTimes:
+            sed = snobj.SNObjectSED(time=time,
+                                    bandpass=self.lsstBandPass['r'])
+            assert not sed.calcADU(bandpass=self.lsstBandPass['r'],
```
danielsf: The project policy is to use the assert methods from ```unittest```, since they give more explicit error messages when they fail.  In this case you could use

```
self.assertGreater(sed.calcADU(...), 0.0)
```

which will test that the result of ```calcADU()``` is greater than 0

danielsf: here is the documentation for the assert methods in ```unittest```

https://docs.python.org/2.7/library/unittest.html#assert-methods

danielsf:
in ```SNObject_example.ipynb```, there are too many instances of ```Sed``` in this sentence

The SN sed at a time mjdobs and an array of wavelengths in units of nm is obtained in the following manner as sims.photUtils.Sed.Sed instance.

danielsf:
Other than the comments already made, this looks good to merge.

examples/testSN.py
rbiswas4: Done

examples/testSN.py
rbiswas4: Done.

rbiswas4:
Regardless I think we should retry moving testSN.py to where it belongs. We apparently are trying to integrate with astropy so this should work sooner or later.

rbiswas4:
Changes Done.

PR #55, opened 2016-06-14
U/danielsf/light curve generator

danielsf:
This pull request implements a prototype LightCurveGenerator which takes an observing cadence represented by an OpSim output database and converts that into light curves for transient and variable sources stored in fatboy (or some other CatSim compatible database).

The supernova and AGN simulation codes were also touched in an effort to speed them up.

### DIFF
--- a/python/lsst/sims/catUtils/mixins/VariabilityMixin.py
+++ b/python/lsst/sims/catUtils/mixins/VariabilityMixin.py
@@ -268,12 +268,7 @@ class Variability(object):
         # use that previous simulation as the starting point.
         #
         if cache_name in _agn_lc_cache:
-            if hasattr(expmjd, '__len__'):
-                first_date = expmjd.min()
-            else:
-                first_date = expmjd
-
-            if _agn_lc_cache[cache_name]['mjd'] <first_date:
+            if _agn_lc_cache[cache_name]['mjd'] <expmjd:
                 resumption = True
 
         if resumption:
@@ -287,10 +282,9 @@ class Variability(object):
             for k in sfint:
                 dx_0[k]=0.0
 
-        epochs = expmjd - start_date
-        endepoch = epochs.max()
+        endepoch = expmjd - start_date
 
-        if epochs.min() < 0:
+        if endepoch < 0:
             raise RuntimeError("WARNING: Time offset greater than minimum epoch.  " +
                                "Not applying variability. "+
                                "expmjd: %e should be > toff: %e  " % (expmjd, toff) +
@@ -298,25 +292,24 @@ class Variability(object):
 
         dt = tau/100.
         nbins = int(math.ceil(endepoch/dt))
-        dt = dt/tau
-        sdt = math.sqrt(dt)
 
-        es = rng.normal(0., 1., nbins)
+        x1 = (nbins-1)*dt
+        x2 = (nbins)*dt
+
+        dt = dt/tau
+        es = rng.normal(0., 1., nbins)*math.sqrt(dt)
         dx_cached = {}
 
         for k in sfint.keys():
-            dx = numpy.zeros(nbins+1)
-            dx[0] = dx_0[k]
+            dx2 = dx_0[k]
             for i in range(nbins):
                 #The second term differs from Zeljko's equation by sqrt(2.)
                 #because he assumes stdev = sfint/sqrt(2)
-                dx[i+1] = -dx[i]*dt + sfint[k]*es[i]*sdt + dx[i]
+                dx1 = dx2
+                dx2 = -dx1*dt + sfint[k]*es[i] + dx1
 
-            dx_cached[k] = dx[nbins]
-            x = numpy.linspace(0, endepoch, nbins+1)
-            intdx = interp1d(x, dx)
-            magoff = intdx(epochs)
-            dMags[k] = magoff
+            dx_cached[k] = dx2
+            dMags[k] = (endepoch*(dx1-dx2)+dx2*x1-dx1*x2)/(x1-x2)
 
         if len(_agn_lc_cache)>1000000:
             _agn_lc_cache = {}
@@ -324,7 +317,7 @@ class Variability(object):
         if cache_name not in _agn_lc_cache:
             _agn_lc_cache[cache_name] = {}
 
-        _agn_lc_cache[cache_name]['mjd'] = start_date+nbins*dt*tau
+        _agn_lc_cache[cache_name]['mjd'] = start_date+x2
         _agn_lc_cache[cache_name]['rng'] = copy.deepcopy(rng)
         _agn_lc_cache[cache_name]['dx'] = dx_cached
 
--- a/python/lsst/sims/catUtils/mixins/VariabilityMixin.py
+++ b/python/lsst/sims/catUtils/mixins/VariabilityMixin.py
@@ -12,12 +12,16 @@ from scipy.interpolate import interp1d
 __all__ = ["Variability", "VariabilityStars", "VariabilityGalaxies",
            "reset_agn_lc_cache"]
 
-_agn_lc_cache = {} # a global cache of agn light curve calculations
+_AGN_LC_CACHE = {} # a global cache of agn light curve calculations
 
 
 def reset_agn_lc_cache():
-    global _agn_lc_cache
-    _agn_lc_cache = {}
+    """
+    Resets the _AGN_LC_CACHE (a global dict for cacheing time steps in AGN
+    light curves) to an empty dict.
+    """
+    global _AGN_LC_CACHE
+    _AGN_LC_CACHE = {}
     return None
 
 
@@ -236,12 +240,11 @@ class Variability(object):
 
 
     @register_method('applyAgn')
-    def applyAgn(self, params, expmjd_in):
+    def applyAgn(self, params, expmjd):
 
-        global _agn_lc_cache
+        global _AGN_LC_CACHE
 
         dMags = {}
-        expmjd = numpy.asarray(expmjd_in,dtype=float)
         toff = numpy.float(params['t0_mjd'])
         seed = int(params['seed'])
         sfint = {}
@@ -256,7 +259,7 @@ class Variability(object):
         # A string made up of this AGNs variability parameters that ought
         # to uniquely identify it.
         #
-        cache_name = '%d_%.12f_%.12f_%.12f_%.12f_%.12f_%.12f_%.12f_%.12f' \
+        agn_ID = '%d_%.12f_%.12f_%.12f_%.12f_%.12f_%.12f_%.12f_%.12f' \
         %(seed, sfint['u'], sfint['g'], sfint['r'], sfint['i'], sfint['z'],
           sfint['y'], tau, toff)
 
@@ -267,14 +270,14 @@ class Variability(object):
         # earlier than the first requested MJD.  If so,
         # use that previous simulation as the starting point.
         #
-        if cache_name in _agn_lc_cache:
-            if _agn_lc_cache[cache_name]['mjd'] <expmjd:
+        if agn_ID in _AGN_LC_CACHE:
+            if _AGN_LC_CACHE[agn_ID]['mjd'] <expmjd:
                 resumption = True
 
         if resumption:
-            rng = copy.deepcopy(_agn_lc_cache[cache_name]['rng'])
-            start_date = _agn_lc_cache[cache_name]['mjd']
-            dx_0 = _agn_lc_cache[cache_name]['dx']
+            rng = copy.deepcopy(_AGN_LC_CACHE[agn_ID]['rng'])
+            start_date = _AGN_LC_CACHE[agn_ID]['mjd']
+            dx_0 = _AGN_LC_CACHE[agn_ID]['dx']
         else:
             start_date = toff
             rng = numpy.random.RandomState(seed)
@@ -311,15 +314,18 @@ class Variability(object):
             dx_cached[k] = dx2
             dMags[k] = (endepoch*(dx1-dx2)+dx2*x1-dx1*x2)/(x1-x2)
 
-        if len(_agn_lc_cache)>1000000:
-            _agn_lc_cache = {}
+        # Reset that AGN light curve cache once it contains
+        # one million objects (to prevent it from taking up
+        # too much memory).
+        if len(_AGN_LC_CACHE)>1000000:
+            reset_agn_lc_cache()
 
-        if cache_name not in _agn_lc_cache:
-            _agn_lc_cache[cache_name] = {}
+        if agn_ID not in _AGN_LC_CACHE:
+            _AGN_LC_CACHE[agn_ID] = {}
 
-        _agn_lc_cache[cache_name]['mjd'] = start_date+x2
-        _agn_lc_cache[cache_name]['rng'] = copy.deepcopy(rng)
-        _agn_lc_cache[cache_name]['dx'] = dx_cached
+        _AGN_LC_CACHE[agn_ID]['mjd'] = start_date+x2
+        _AGN_LC_CACHE[agn_ID]['rng'] = copy.deepcopy(rng)
+        _AGN_LC_CACHE[agn_ID]['dx'] = dx_cached
 
         return dMags
 
--- a/python/lsst/sims/catUtils/mixins/sncat.py
+++ b/python/lsst/sims/catUtils/mixins/sncat.py
@@ -53,7 +53,7 @@ class SNFunctionality(object):
     # 'mag_u', 'mag_g', 'mag_r', 'mag_i', 'mag_z', 'mag_y']
     cannot_be_null = ['x0', 'z', 't0']
 
-    @astropy.utils.lazyproperty
+    @property
     def mjdobs(self):
         '''
         The time of observation for the catalog, which is set to be equal
--- a/python/lsst/sims/catUtils/mixins/sncat.py
+++ b/python/lsst/sims/catUtils/mixins/sncat.py
@@ -194,7 +194,7 @@ class SNFunctionality(object):
         vals = np.array([[0.0]*len(t0), [np.inf]*len(t0),
                         [np.nan]*len(t0), [np.inf]*len(t0)]).transpose()
 
-        for i in np.where(np.isfinite(t0))[0]:
+        for i in np.where(np.logical_and(np.isfinite(t0), np.abs(self.mjdobs-t0)<self.maxTimeSNVisible))[0]:
             SNobject.set(z=_z[i], c=c[i], x1=x1[i], t0=t0[i], x0=x0[i])
             if self.mjdobs<=SNobject.maxtime() and self.mjdobs>=SNobject.mintime():
                 SNobject.setCoords(ra=raDeg[i], dec=decDeg[i])
--- a/python/lsst/sims/catUtils/mixins/sncat.py
+++ b/python/lsst/sims/catUtils/mixins/sncat.py
@@ -179,33 +179,34 @@ class SNFunctionality(object):
 
         for i in np.where(np.isfinite(t0))[0]:
             SNobject.set(z=_z[i], c=c[i], x1=x1[i], t0=t0[i], x0=x0[i])
-            SNobject.setCoords(ra=ra[i], dec=dec[i])
-            SNobject.mwEBVfromMaps()
+            if self.mjdobs<=SNobject.maxtime() and self.mjdobs>=SNobject.mintime():
+                SNobject.setCoords(ra=ra[i], dec=dec[i])
+                SNobject.mwEBVfromMaps()
 
-            # Calculate fluxes
-            fluxinMaggies = SNobject.catsimBandFlux(time=self.mjdobs,
-                                                    bandpassobject=bandpass)
-            mag = SNobject.catsimBandMag(time=self.mjdobs,
-                                         fluxinMaggies=fluxinMaggies,
-                                         bandpassobject=bandpass)
-            vals[i, 0] = fluxinMaggies
-            vals[i, 1] = mag
-            flux_err = SNobject.catsimBandFluxError(time=self.mjdobs,
-                                                    bandpassobject=bandpass,
-                                                    m5=self.obs_metadata.m5[
-                                                        bandname],
-                                                    photParams=None,
-                                                    fluxinMaggies=fluxinMaggies,
-                                                    magnitude=mag)
+                # Calculate fluxes
+                fluxinMaggies = SNobject.catsimBandFlux(time=self.mjdobs,
+                                                        bandpassobject=bandpass)
+                mag = SNobject.catsimBandMag(time=self.mjdobs,
+                                             fluxinMaggies=fluxinMaggies,
+                                             bandpassobject=bandpass)
+                vals[i, 0] = fluxinMaggies
+                vals[i, 1] = mag
+                flux_err = SNobject.catsimBandFluxError(time=self.mjdobs,
+                                                        bandpassobject=bandpass,
+                                                        m5=self.obs_metadata.m5[
+                                                            bandname],
+                                                        photParams=None,
+                                                        fluxinMaggies=fluxinMaggies,
+                                                        magnitude=mag)
 
-            mag_err = SNobject.catsimBandMagError(time=self.mjdobs,
-                                                  bandpassobject=bandpass,
-                                                  m5=self.obs_metadata.m5[
-                                                      bandname],
-                                                  photParams=None,
-                                                  magnitude=mag)
-            vals[i, 2] = flux_err
-            vals[i, 3] = mag_err
+                mag_err = SNobject.catsimBandMagError(time=self.mjdobs,
+                                                      bandpassobject=bandpass,
+                                                      m5=self.obs_metadata.m5[
+                                                          bandname],
+                                                      photParams=None,
+                                                      magnitude=mag)
+                vals[i, 2] = flux_err
+                vals[i, 3] = mag_err
 
         return (vals[:, 0], vals[:, 1], vals[:, 2], vals[:, 3])
 
--- a/python/lsst/sims/catUtils/mixins/sncat.py
+++ b/python/lsst/sims/catUtils/mixins/sncat.py
@@ -140,12 +140,15 @@ class SNFunctionality(object):
             self.column_by_name('raJ2000'),\
             self.column_by_name('decJ2000')
 
+        raDeg = np.degrees(ra)
+        decDeg = np.degrees(dec)
+
         SNobject = SNObject()
 
         sedlist = []
         for i in range(self.numobjs):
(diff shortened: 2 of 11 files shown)


python/lsst/sims/catUtils/mixins/VariabilityMixin.py
```diff
@@ -2,13 +2,24 @@
 import linecache
```
rbiswas4: It does not look like this package is being used. So maybe the import can be dropped.

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
```diff
@@ -2,13 +2,24 @@
...
 import math
 import os
+import copy
 import json as json
 from lsst.sims.catalogs.measures.instance import register_class, register_method, compound
 from scipy.interpolate import InterpolatedUnivariateSpline
 from scipy.interpolate import UnivariateSpline
```
rbiswas4: Unused package. I think you may have dropped this in favor of InterpolatedUnivariateSpline

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
```diff
@@ -2,13 +2,24 @@
...
 from scipy.interpolate import interp1d
 
-__all__ = ["Variability", "VariabilityStars", "VariabilityGalaxies"]
+__all__ = ["Variability", "VariabilityStars", "VariabilityGalaxies",
+           "reset_agn_lc_cache"]
+
+_agn_lc_cache = {} # a global cache of agn light curve calculations
+
+
+def reset_agn_lc_cache():
```
rbiswas4: Public function should have a docstring. The variable _agn_lc_cache should probably be all Uppercase according to conventions, and starting with the underscore as done.

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
```diff
@@ -238,32 +252,75 @@ def applyAgn(self, params, expmjd_in):
...
+        else:
+            start_date = toff
+            rng = numpy.random.RandomState(seed)
+            dx_0 = {}
+            for k in sfint:
+                dx_0[k]=0.0
+
+        endepoch = expmjd - start_date
+
+        if endepoch < 0:
```
rbiswas4: `RuntimeError` seems like an exception and cause the program to crash. Does it make sense to have a message that seems to warn the user that  variability will not be applied?

I think you do not need the '+' signs. Within a parenthesis, the strings will be joined.

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
```diff
@@ -226,6 +237,9 @@ def applyMicrolens(self, params, expmjd_in):
 
     @register_method('applyAgn')
     def applyAgn(self, params, expmjd_in):
```
rbiswas4: Missing docstring.

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
```diff
@@ -238,32 +252,75 @@ def applyAgn(self, params, expmjd_in):
         sfint['z'] = params['agn_sfz']
         sfint['y'] = params['agn_sfy']
         tau = params['agn_tau']
-        epochs = expmjd - toff
-        if epochs.min() < 0:
+
+        # A string made up of this AGNs variability parameters that ought
+        # to uniquely identify it.
+        #
```
rbiswas4: The line 258 is unnecessary. Would something like agn_ID  be a better name for readability.

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
```diff
@@ -238,32 +252,75 @@ def applyAgn(self, params, expmjd_in):
...
+
+        resumption = False
+
+        # Check to see if this AGN has already been simulated.
+        # If it has, see if the previously simulated MJD is
+        # earlier than the first requested MJD.  If so,
+        # use that previous simulation as the starting point.
+        #
+        if cache_name in _agn_lc_cache:
+            if _agn_lc_cache[cache_name]['mjd'] <expmjd:
```
rbiswas4: It seemed like expmjd is an array. This might work if the array has length 1, but it might be better to make it a float, or use all, any etc.

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
```diff
@@ -238,32 +252,75 @@ def applyAgn(self, params, expmjd_in):
...
-        dt = (endepoch/nbins)/tau
-        sdt = math.sqrt(dt)
-        numpy.random.seed(seed=seed)
-        es = numpy.random.normal(0., 1., nbins)
+
+        x1 = (nbins-1)*dt
+        x2 = (nbins)*dt
+
+        dt = dt/tau
+        es = rng.normal(0., 1., nbins)*math.sqrt(dt)
```
rbiswas4: So, this is an array of random numbers drawn from the normal distribution for each bin and divided by 10. independent of any physical paramters?

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
```diff
@@ -238,32 +252,75 @@ def applyAgn(self, params, expmjd_in):
...
-            x = numpy.linspace(0, endepoch, nbins+1)
-            intdx = interp1d(x, dx)
-            magoff = intdx(epochs)
-            dMags[k] = magoff
+                dx1 = dx2
+                dx2 = -dx1*dt + sfint[k]*es[i] + dx1
+
+            dx_cached[k] = dx2
+            dMags[k] = (endepoch*(dx1-dx2)+dx2*x1-dx1*x2)/(x1-x2)
+
```
rbiswas4: Add a comment on why the cache is being reset. Is this physical, or a space saving device.

python/lsst/sims/catUtils/supernovae/snObject.py
```diff
@@ -520,7 +524,20 @@ def SNObjectSED(self, time, wavelen=None, bandpass=None,
             return SEDfromSNcosmo
 
         # Apply LSST extinction
-        ax, bx = SEDfromSNcosmo.setupCCMab()
+        global _sn_ax_cache
+        global _sn_bx_cache
+        global _sn_ax_bx_wavelen
+        if _sn_ax_bx_wavelen is None \
+        or len(wavelen)!=len(_sn_ax_bx_wavelen) \
```
rbiswas4: Is this length comparison not already taken care of by the following element by element comparison?

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -166,44 +183,60 @@ def get_snbrightness(self):
...
+        for i in np.where(np.logical_and(np.isfinite(t0), np.abs(self.mjdobs-t0)<self.maxTimeSNVisible))[0]:
+
+            if id_list[i] in self._sn_object_cache:
+                SNobject = self._sn_object_cache[id_list[i]]
+            else:
+                SNobject = SNObject()
+                SNobject.set(z=_z[i], c=c[i], x1=x1[i], t0=t0[i], x0=x0[i])
+                SNobject.setCoords(ra=raDeg[i], dec=decDeg[i])
+                SNobject.set_MWebv(ebv[i])
+                self._sn_object_cache[id_list[i]] = SNobject
```
rbiswas4: Just recording this as something to try later:

Instead of Caching SNObject instances, we could either keep pickled objects or SNObject.SNstate in some format. Currently SNstate is a dictionary and maybe changing it to a named tuple will decrease the size of the variable. The point is that SNObject instances can be reconstructed from SNstate via the 
```
snstate = snobject.SNstate # snobject is an instance of SNObject
s = SNObject.fromSNstate(snstate)
# s should have the same properties as snobject
```
This was the kind of purpose, these methods were built for.

python/lsst/sims/catUtils/utils/ObservationMetaDataGenerator.py
```diff
@@ -221,6 +222,8 @@ def getObservationMetaData(self, obsHistID=None, expDate=None, fieldRA=None, fie
 
                 nConstraints += 1
 
+        query += ' GROUP BY expMJD'
```
rbiswas4: Given the causality requirements of the AGN simulation, have you considered adding and ` ORDER BY expMJD` as well? Or does that slow doe the generation a lot?

python/lsst/sims/catUtils/utils/LightCurveGenerator.py
```diff
@@ -0,0 +1,496 @@
...
+        ----------
+        chunk_size : int, optional, defaults to None
+            the number of rows to return from the database at a time. If None,
+            returns the entire database query in one chunk.
+
+        query_cache : iterator over database rows, optional, defaults to None
+            the result of calling db_obj.query_columns().  If query_cache is not
+            None, this method will iterate over the rows in query_cache and produce
+            an appropriate InstanceCatalog. DO NOT use this
+            unless you know what you are doing.  It is an optional
```
rbiswas4: - Is it useful to use the _baseLightCurve and not use query_cache? Maybe the use of custom LightCurveGenerator classes (as opposed to those already provided by catsim) should be discouraged rather than a single variable. 
- Even if the above is not true, DO NOT use should probably be replaced by something saying that non None values should not be used.

python/lsst/sims/catUtils/utils/LightCurveGenerator.py
```diff
@@ -0,0 +1,496 @@
...
+        else:
+            obs_list = []
+            for bp in bandpass:
+                sub_list = self._generator.getObservationMetaData(fieldRA=ra,
+                                                                  fieldDec=dec,
+                                                                  telescopeFilter=bp,
+                                                                  expMJD=expMJD,
+                                                                  boundLength=1.75)
+                obs_list += sub_list
+
```
rbiswas4: Would be nice to add a comment saying obs_list is a list of ObservationMetaData Instances with pointingRA, pointingDec, filters, observed MJD values in the ranges indicated by the tuples ra, dec, expMJD and bp

python/lsst/sims/catUtils/utils/LightCurveGenerator.py
```diff
@@ -0,0 +1,496 @@
...
+                         # in obs_list.  All of the ObservationMetaData in
+                         # obs_list[i] will point to the same point on the sky.
+
+        mjd_groups = []  # a list of lists of the MJDs of the ObservationMetaDatas
+                         # so that they can be sorted into chronological order before
+                         # light curves are calculated
+
+        for iobs, obs in enumerate(obs_list):
+            group_dex = -1
+
```
rbiswas4: Unless this is done over a small patch of the sky, this would be very slow because of the nested loop. On the other hand, to do exactly what is being done right now (no support for dithers and overlaps), one could actually group by fieldIDs. This would be much faster. Even to do the full, computation, this may be a good first step (or replacing this by some tree).

python/lsst/sims/catUtils/utils/LightCurveGenerator.py
```diff
@@ -0,0 +1,496 @@
...
+    def _get_query_from_group(self, grp, chunk_size):
+        """
+        Take a group of ObervationMetaData that all point to the same region
+        of the sky.  Query the CatSim database for all of the celestial objects
+        in that region, and return it as an iterator over database rows.
+        """
+
+        cat = self._lightCurveCatalogClass(self._catalogdb, obs_metadata=grp[0])
+
```
rbiswas4: Is this line on db_required_columns() necessary?

python/lsst/sims/catUtils/utils/LightCurveGenerator.py
```diff
@@ -0,0 +1,496 @@
...
+        grp.
+
+        Output
+        ------
+        This method does not output anything.  It adds light curves to the
+        instance member variables self.mjd_dict, self.mag_dict, self.sig_dict,
+        and self.truth_dict.
+        """
+
+        global _sed_cache
```
rbiswas4: If local_gamma_cache is a cache storing gamma values for each pointing group (fieldID), it might be good to clarify this in a comment.

python/lsst/sims/catUtils/utils/LightCurveGenerator.py
```diff
@@ -0,0 +1,496 @@
...
+
+                # we must sort the MJDs because, if an object appears in multiple
+                # spatial pointings, its observations will be concatenated out of order
+                mjd_arr = np.array(self.mjd_dict[unique_id][bp])
+                mjd_dexes = np.argsort(mjd_arr)
+
+                output_dict[unique_id][bp]['mjd'] = mjd_arr[mjd_dexes]
+                output_dict[unique_id][bp]['mag'] = np.array(self.mag_dict[unique_id][bp])[mjd_dexes]
+                output_dict[unique_id][bp]['error'] = np.array(self.sig_dict[unique_id][bp])[mjd_dexes]
+
```
rbiswas4: Did you intend to leave the print statements in the file for future development or was this unintended?

python/lsst/sims/catUtils/utils/SNIaLightCurveGenerator.py
```diff
@@ -0,0 +1,164 @@
...
+
+class _sniaLightCurveCatalog(_baseLightCurveCatalog, SNIaCatalog, PhotometryBase):
+
+    column_outputs = ["uniqueId", "snid", "raJ2000", "decJ2000",
+                      "cosmologicalDistanceModulus", "redshift", "EBV"]
+
+    _suppressDimSN = False
+
+
+class SNIaLightCurveGenerator(LightCurveGenerator):
```
rbiswas4: Missing docstring. What should go in args, and kwargs at the minimum?

python/lsst/sims/catUtils/utils/SNIaLightCurveGenerator.py
```diff
@@ -0,0 +1,164 @@
...
+    def _light_curves_from_query(self, cat_dict, query_result, grp):
+
+        t_dict = {}
+        gamma_dict = {}
+        m5_dict = {}
+        t_min = None
+        t_max = None
+        for bp_name in cat_dict:
+            self.lsstBandpassDict[bp_name].sbTophi()
+            t_dict[bp_name] = np.array([obs.mjd.TAI for obs in grp
```
rbiswas4: It seems that in these three loops through grp, we are filtering by all cases where the bandpass is bp_name, so maybe we can group these three loops into a single one

python/lsst/sims/catUtils/utils/SNIaLightCurveGenerator.py
```diff
@@ -0,0 +1,164 @@
...
+        snobj = SNObject()
+
+        cat = cat_dict[cat_dict.keys()[0]]  # does not need to be associated with a bandpass
+
+        for chunk in query_result:
+            t_start_chunk = time.time()
+            for sn in cat.iter_catalog(query_cache=[chunk]):
+                sn_rng = self.sn_universe.getSN_rng(sn[1])
+                sn_t0 = self.sn_universe.drawFromT0Dist(sn_rng)
+                if sn[5] <= self.z_cutoff and np.isfinite(sn_t0) and \
```
rbiswas4: While this is OK. There is a mechanism in `sims_catUtils.SNFunctionality` which is inherited by SNIaCatalogs to do this already:
If `maxz` is set to a desired value, and `suppressHighzSN` is set to True. (Defaults are 1.2, True), the equivalent of implementing the zcutoff condition here is implemented. Similarly, if `suppressDimSN` is set to True, the sn_t0 range in the next two lines are implemented.

python/lsst/sims/catUtils/utils/SNIaLightCurveGenerator.py
```diff
@@ -0,0 +1,164 @@
...
+                            m5_active = m5_list[active_dexes]
+                            gamma_active = gamma_list[active_dexes]
+
+                            if len(t_active) > 0:
+                                wave_ang = bandpass.wavelen*10.0
+                                mask = np.logical_and(wave_ang > snobj.minwave(),
+                                                      wave_ang < snobj.maxwave())
+
+                                wave_ang = wave_ang[mask]
+                                snobj.set(mwebv=sn[6])
```
rbiswas4: If you want to use the `sims` extinction maps and methods, you should not use this, but use 
`snobj.set_MWebv(sn[6])`

python/lsst/sims/catUtils/utils/SNIaLightCurveGenerator.py
```diff
@@ -0,0 +1,164 @@
...
+
+                                ss = Sed()
+                                fnu_grid = flambda_grid*bandpass.wavelen* \
+                                           bandpass.wavelen*ss._physParams.nm2m* \
+                                           ss._physParams.ergsetc2jansky/ss._physParams.lightspeed
+
+                                flux_list = \
+                                (fnu_grid*bandpass.phi).sum(axis=1)*(bandpass.wavelen[1]-bandpass.wavelen[0])
+
+                                flux_list = np.array(flux_list)
```
rbiswas4: Why is flux_list not an array already? I assume this is cgs rather than in maggies.

python/lsst/sims/catUtils/utils/SNIaLightCurveGenerator.py
```diff
@@ -0,0 +1,164 @@
...
+                                        self.sig_dict[sn[0]][bp_name] = []
+
+                                for tt, mm, ee in zip(t_active[acceptable], mag_list[acceptable],
+                                                      mag_error_list[0]):
+
+                                    self.mjd_dict[sn[0]][bp_name].append(tt)
+                                    self.mag_dict[sn[0]][bp_name].append(mm)
+                                    self.sig_dict[sn[0]][bp_name].append(ee)
+
+            print("chunk of ", len(chunk), " took ", time.time()-t_start_chunk)
```
rbiswas4: Does this mean that flux and flux errror  which are the quantities of interest for SN, are not returned?

python/lsst/sims/catUtils/utils/SNIaLightCurveGenerator.py
```diff
@@ -0,0 +1,164 @@
...
+                                for ff, ff_sn in zip(flambda_grid, sn_ff_buffer):
+                                    ff[mask] = np.where(ff_sn > 0.0, ff_sn, 0.0)
+
+                                ss = Sed()
+                                fnu_grid = flambda_grid*bandpass.wavelen* \
+                                           bandpass.wavelen*ss._physParams.nm2m* \
+                                           ss._physParams.ergsetc2jansky/ss._physParams.lightspeed
+
+                                flux_list = \
+                                (fnu_grid*bandpass.phi).sum(axis=1)*(bandpass.wavelen[1]-bandpass.wavelen[0])
```
rbiswas4: It might be faster to use `np.dot`, I often find that to be faster than `np.sum()`.  Also the $d\lambda$  term is probably `bandpass.step`. You should be able to do the calculation from flambda_grid to fluxes/magnitudes directly without going to fnu.

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
danielsf: copy is used to cache the random number generator (there are calls to copy.deepcopy() in the AGN caching code)

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
danielsf: Quirky observation: you do need the '+' signs once you start concatenating strings of the form

"text and %d" % my_variable

Given that this mixin will only be used to add variability to an AGN, it seems appropriate to me that the code should crash if variability is impossible (if the user wanted a static AGN, they could have made an InstanceCatalog that did not inherit from VariabilityGalaxies).  I am going to leave it for now.  We can discuss more later, if we want (there are a bunch of exceptions that are poorly handled in the Variability code).  It is a design decision that was made a while ago, and I am erring on the side of not doing anything.

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
danielsf: None of the 'apply[SpecificVariabilityMethod]' methods have docstrings.  This is clearly a problem, but they're also not (in my opinion) designed to be easily called outside of the "pass a varParamStr to applyVariability()" model (i.e. they all just take a dict 'params' without specifying its contents).  I think we should make another ticket to refactor these methods into something that can be called from applyVariability() and something that can be called by hand outside of InstanceCatalogs.

danielsf: see SIM-2038

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
danielsf: Good catch.  The method used to be vectorized in expmjd, but that did not make sense in the context of the caching (which I think gets us more than the vectorization, since the vectorization still involved nested for loops).  I was trying to de-vectorize it.  I will revert expmjd to a float.

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
danielsf: Yes.  I believe that is how the model works, though I have not been able to satisfactorily track down the reference we are using (I think it is ApJ 721:1014).

Note: it is not divided by 10.  I think you read 1. as 0.1

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
danielsf: It is a memory-saving issue.

python/lsst/sims/catUtils/supernovae/snObject.py
danielsf: The element-by-element comparison raises an exception if the lengths are different (so: "no")

python/lsst/sims/catUtils/utils/ObservationMetaDataGenerator.py
danielsf: Yeah.  I have yet to see things come back out of chronological order, but I haven't verified that that is implied by GROUP BY

python/lsst/sims/catUtils/utils/LightCurveGenerator.py
danielsf: I agree: only advanced users should try to write their own LightCurveGenerators.  I guess I wanted the API to look as much like the "normal" iter_catalog() as possible so that those advanced users (who are presumably familiar with InstanceCatalog) can intuit what is going on.

I will change the docstring as you suggest.

python/lsst/sims/catUtils/utils/LightCurveGenerator.py
danielsf: Let's do some profiling of this.  During the brainstorming session before I left, Peter suggested that we show users how to draw up their own mock-OpSim databases so that they can test personalized observing cadences.  I would like to minimize the number of OpSim columns that they have to emulate to pull that off.

python/lsst/sims/catUtils/utils/LightCurveGenerator.py
danielsf: I believe so.  That is where InstanceCatalog verifies that your CatalogDBObject class contains all of the raw columns needed to generate your getter-calculated columns.

python/lsst/sims/catUtils/utils/LightCurveGenerator.py
danielsf: Yeah... or just to give users a heads up if things are going slower than they intended.  I will eventually take them out.

python/lsst/sims/catUtils/utils/SNIaLightCurveGenerator.py
danielsf: done

python/lsst/sims/catUtils/utils/SNIaLightCurveGenerator.py
danielsf: It looks like suppressHighzSN will set z=np.nan if suppressDimSN is set to True.  However, it is not clear to me that, if z=np.nan, drawSNParams (and, more importantly, drawFromX0Dist, since that is what spends the most time) is not called (i.e. even if we suppress the high z supernovae, I *think* sncat.py still spends time calculating x0 values for the supernovae, which is what I am trying to avoid).

python/lsst/sims/catUtils/utils/SNIaLightCurveGenerator.py
danielsf: ---
If you want to use the sims extinction maps and methods, you should not use this, but use
snobj.set_MWebv(sn[6]) Added a ticket for myself to prevent accidental usage of this kind
---

Why?  The unittest validates the light curve generator against your InstanceCatalogs.  They agree at the sub-millimag level.

danielsf: I caught the "if you ask sncosmo for fluxes outside of its wavelength range, it crashes" problem.  I tried to just copy your code from sncat.py here to deal with that.

python/lsst/sims/catUtils/utils/SNIaLightCurveGenerator.py
danielsf: the cast to np.array was left over from an earlier attempt to make the code work.

yes: it should be cgs flux units

python/lsst/sims/catUtils/utils/SNIaLightCurveGenerator.py
rbiswas4: Ok. That sounds like a bug in the sense of unnecessary time spent in sncat.py. Maybe there should be a ticket to check that and change it there as well so that this is not happening.

python/lsst/sims/catUtils/utils/SNIaLightCurveGenerator.py
rbiswas4: 'Why? The unittest validates the light curve generator against your InstanceCatalogs. They agree at the sub-millimag level.'

- Because having multiple methods to do the same thing in different parts of our codebase would make life really hard for usual reasons of avoiding code duplication. Essentially, I want all bug-fixes and improvements in extinction to be applied in the photUtils method, and all SN extinctions related issues to be in one place as far as possible.
- I know you caught the rectification issue, but I did not notice if you caught the case where the non-rectified SED has the outside the wavelength range problem.
- I would like to add a method to prevent accidental usage as I see this as something an user is likely to try (in analogy with other set methods) without noticing thee things like rectification etc. (which you did).

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
rbiswas4: Sounds good .... While reviewing this I was also noticing sncat is missing some docstrings, maybe this should be added to the ticket as well.

PR #56, opened 2016-06-15
Sim 2029 redshifting

rbiswas4:
This is a PR for https://jira.lsstcorp.org/browse/SIM-2029. 

Previous behavior for SNObject is preserved, but a new method has been added as a feature. The feature method is a method called 
```
def redshift(self, z, cosmo)
```
which can be used to change the redshift to z for a certain cosmology. 

- This is illustrated in `examples/SNObject_example.py`
- Tests have been added to `examples/testSN.py` to perform sanity checks
- Jenkins builds pass : https://ci.lsst.codes/job/stack-os-matrix/12321/


python/lsst/sims/catUtils/supernovae/snObject.py
```diff
@@ -420,6 +419,49 @@ def mwEBVfromMaps(self):
...
+        Returns
+        -------
+        None, but it changes the instance
+
+        """
+        import types
+
+        # Check that the input redshift is a scalar
+        try:
+            assert isinstance(z, types.FloatType) or \
```
danielsf: something I learned last week:  you can
```
import numbers
isinstance(x, numbers.Number)
```
to test if x is an int or a float

rbiswas4: That would definitely be a better one. I have used this before but was having a hard time finding the exact syntax to do it.

danielsf:
You asked me to take a look at this.  I think it looks fine.  If @jbkalmbach doesn't have time to look at it, I would say that it is ready to merge.

jbkalmbach:
Looks good. Only suggestion is that in your example you show the relevant magnitudes changing/not changing as appropriate since those are the actual observables.

PR #57, opened 2016-06-28
Split observation meta data generatory

rbiswas4:
Split functionality on ObservationMetaDataGenerator into :
- getOpSimRecords : obtain OpSim records matching the intersection of user
    specified ranges on each column in the OpSim output database. The
    records are in the form of a `numpy.recarray`
- ObservationMetaDataForPointing : convert an OpSim record for a single
    OpSim Pointing to an instance of ObservationMetaData usable by catsim
    and PhoSim Instance Catalogs.
- getObservationMetaData : Obtain a list of ObservationMetaData instances
    corresponding to OpSim pointings matching the intersection of user
     specified ranges on each column in the OpSim output database. This should
     be identitcal to the old `getObservationMetaData`

The old method  `getObservationMetaData` is still there as `getObservationMetaDataOld` only for testing purposes.

A few tests were added to check that the OpSimRecords queried had the same properties as the ObservationMetaDataInstances. 

Example Notebook added for demonstration of usage where the OpSim records have been selected by the user.

### DIFF
--- a/examples/ObservationMetaData_Generator_Example.ipynb
+++ b/examples/ObservationMetaData_Generator_Example.ipynb
@@ -1,0 +1,489 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "import numpy"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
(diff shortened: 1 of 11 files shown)


danielsf:
I used git cherry-pick

https://git-scm.com/docs/git-cherry-pick

to create a branch u/danielsf/SIM-1794/cherry_pick that only contains the commits that should have been in this pull request.  I will close this pull request and open a new pull request on that branch.  That is where I will do the review.

PR #61, opened 2016-07-08
Feature/sim 1471/pho sim meta data

danielsf:
This pull request re-factors the way we handle PhoSim header information in ObservationMetaData.  Instead of taking all of the header information from ObservationMetaData.phoSimMetaData, pointing information, date, rotSkyPos, and bandpass are taken from the appropriate ObservationMetaData member variables.  phoSimMetaData only exists to store additional information that may have come from an Opsim pointing (or from user input).

A method write_phoSim_header() is added to the PhoSim InstanceCatalog classes in sims_catUtils to parse this new API.

ObservationMetaDataGenerator.py is also updated.

There are pull requests in

sims_utils
sims_catalogs_generation
sims_catUtils
sims_GalSimInterface

### DIFF
--- a/python/lsst/sims/catUtils/utils/ObservationMetaDataGenerator.py
+++ b/python/lsst/sims/catUtils/utils/ObservationMetaDataGenerator.py
@@ -86,6 +86,9 @@ class ObservationMetaDataGenerator(object):
         # to the OpSim database (i.e. OpSim stores all angles in radians; we would like users to be
         # able to specify angles in degrees)
         #
+        # The 5th column is any coordinate transformation required to go from the OpSim database
+        # to the PhoSim header
+        #
         # Note that this conforms to an older
         # PhoSim API.  At some time in the future, both this and OpSim3_61DBObject.py
         # (and possibly phoSimCatalogExamples.py) will need to be updated to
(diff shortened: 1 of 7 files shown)


python/lsst/sims/catUtils/utils/ObservationMetaDataGenerator.py
```diff
@@ -63,88 +120,26 @@ def __init__(self, database, driver='sqlite', host=None, port=None):
...
+        for column in self._user_interface_to_opsim:
+            rec = self._user_interface_to_opsim[column]
+            if rec[0] in summary_columns:
                 self.active_columns.append(column)
-                dtypeList.append((column[1],column[3]))
+                dtypeList.append((rec[0], rec[2]))
                 if self.baseQuery != 'SELECT':
                     self.baseQuery += ','
-                self.baseQuery += ' ' + column[1]
+                self.baseQuery += ' ' + rec[0]
```
rbiswas4: When might this conditional fail? Are there keys in `self._user_interface_to_opsim` that do not exist `summary_columns`? 

Possibly, this might happen for some truncated OpSim or OpSim like object someone has created. Is just not doing anything for missing keys ok?

python/lsst/sims/catUtils/utils/ObservationMetaDataGenerator.py
```diff
@@ -193,119 +188,17 @@ def getOpSimRecords(self, obsHistID=None, expDate=None, fieldRA=None,
...
-
-        @param [in] visitExpTime the exposure time in seconds
-        @param [in] obsHistID the integer used by OpSim to label pointings
-        @param [in] expDate is the date of the exposure (units????)
-        @param [in] expMJD is the MJD of the exposure
-        @param [in] m5 is the five sigma depth of the observation
-        @param [in] skyBrightness
-        """
+        summary_columns = self.opsimdb.get_column_names('Summary')
+        self._set_seeing_column(summary_columns)
```
rbiswas4: Does it make sense to repeat this calculation after __init__ . 
- Can we keep summary_columns as an attribute of the class so that instance methods like this can use it.
- Also, the `__init__` method already set the seeing column, so this is probably unnecessary here ?

python/lsst/sims/catUtils/utils/ObservationMetaDataGenerator.py
```diff
@@ -63,88 +120,26 @@ def __init__(self, database, driver='sqlite', host=None, port=None):
...
-        # Just use the static method to assign this attribute
-        # to the mapping of names of phoSim to OpSim columns
-        self.columnMapping = self.OpSimColumnMap(self._seeing_column)
-
-        #Set up self.dtype containg the dtype of the recarray we expect back from the SQL query.
-        #Also setup baseQuery which is just the SELECT clause of the SQL query
-        #
-        #self.active_columns will be a list containing the subset of columnMapping columns
-        #that actually exist in this opsim database
+        # self.active_columns will be a list containing the subset of columnMapping columns
```
rbiswas4: `columnMapping` -> `_user_interface_to_opsim` probably

python/lsst/sims/catUtils/utils/ObservationMetaDataGenerator.py
```diff
@@ -55,6 +68,50 @@ def __init__(self, database, driver='sqlite', host=None, port=None):
         self.host = host
         self.port = port
         self.database = database
+        self._seeing_column = 'FWHMeff'
+
+        # a dict keyed on the user interface (i.e. args to getObservationMetaData)
+        # names of OpSim data columns.  Returns a tuple that is the
+        # (name of data in OpSim, transformation to go from user interface to OpSim units, dtyp in OpSim)
+        self._user_interface_to_opsim = {'obsHistID': ('obsHistID', None, np.int64),
```
rbiswas4: This is a pretty fundamental attribute containing hard coded information about the mapping of OpSim variables to the user interface. It might make sense to make this a more 'public' variable.

python/lsst/sims/catUtils/utils/ObservationMetaDataGenerator.py
```diff
@@ -193,119 +188,17 @@ def getOpSimRecords(self, obsHistID=None, expDate=None, fieldRA=None,
...
 
         query = self.baseQuery + ' FROM SUMMARY'
 
         nConstraints = 0  # the number of constraints in this query
 
-        for column in self.columnMapping:
-            value = eval(column[0])
+        for column in self._user_interface_to_opsim:
+            transform = self._user_interface_to_opsim[column]
+            value = eval(column)
```
rbiswas4: A comment explaining that you are obtaining the values of the parameter to this function where the parameter has the same name as the key in the `_user_interface_to_opsim` would be nice.

python/lsst/sims/catUtils/utils/ObservationMetaDataGenerator.py
```diff
@@ -314,27 +207,27 @@ def getObservationMetaDataOld(self, obsHistID=None, expDate=None, fieldRA=None,
...
-                    if column[4] is not None:
-                        vv = column[4](value)
+                    if transform[1] is not None:
+                        vv = transform[1](value)
                     else:
                         vv = value
-                    query += ' %s == %s' % (column[1], vv)
+                    query += ' %s == %s' % (transform[0], vv)
 
                 nConstraints += 1
```
rbiswas4: Why are observations being grouped by expMJD ? In OpSim outputs, there are no two observations with the same expMJD (once you purge of duplicates). In principle, (say if you have a beam splitter), one might have two observations with the same expMJD. Will this still work for that case? 

Is this basically trying to order observations? If so can we use ` ORDER BY expMJD`?

python/lsst/sims/catUtils/utils/ObservationMetaDataGenerator.py
danielsf: You are correct: I put this here so if people wanted to create their own "mock OpSim" cadence that did not contain all of the columns in the true OpSim summary table, the ```ObservationMetaDataGenerator``` would still work.

I think ignoring missing columns is the correct thing to do.  I don't want to try to calculating things (for example ```fiveSigmaDepth```) invisibly, in case users have their own models they want to use for missing values.

python/lsst/sims/catUtils/utils/ObservationMetaDataGenerator.py
danielsf: I repeated the ```_set_seeing_column()``` calculation because the new ```ObservationMetaDataFromPointing``` methods mean that users could, in principle, connect to an OpSimDB that using ```finSeeing``` but then pass OpSim records from a newer database that uses ```FWHMeff``` into ```ObservationMetaDataFromPointing()```.

We can probably get away with just setting ```summary_columns``` once, since there are no methods to connect an ```ObservationMetaDataGenerator``` to a new OpSimDB after instantiation.

python/lsst/sims/catUtils/utils/ObservationMetaDataGenerator.py
danielsf: I don't want users to be able to change this, since it defines the expected OpSim schema and I think it should be our responsibility to keep that up-to-date.

Users should only care about this if they are trying to draw up their own mock OpSim cadence, in which case I think they should consult the OpSim documentation to learn what columns are contained in the Summary table.

python/lsst/sims/catUtils/utils/ObservationMetaDataGenerator.py
danielsf: Good point about active_columns.  I think the correct thing to do would be to fail, telling the user that they asked for a constraint on a column that does not exist (so that users don't believe that they selected on something which was actually ignored).  I will add some code that verifies that ```column``` is in ```self.active_columns``` and raises an exception if it is not.

python/lsst/sims/catUtils/utils/ObservationMetaDataGenerator.py
rbiswas4: Yes, trying to recalculate those columns probably does not make sense. I was wondering if missing values merit an exception or warning for people to figure out why an obsmetadata did not come out the way they expected..

python/lsst/sims/catUtils/utils/ObservationMetaDataGenerator.py
rbiswas4: OK. I was more worried about being able to understand what is going on. But, I agree with keeping this as a protected variable ... maybe having some more documentation about this is a better way to address that  than trying to make this public.

python/lsst/sims/catUtils/utils/ObservationMetaDataGenerator.py
danielsf: The group by is there to trim out the duplicates, as per the recommendations here

https://www.lsst.org/scientists/simulations/opsim/summary-table-column-descriptions-v335

PR #62, opened 2016-07-10
Sncosmo v1.3.0 and tests

rbiswas4:
- Added tests to check the import and version of sncosmo and astropy
- passes builds:  https://ci.lsst.codes/job/stack-os-matrix/13092/

### DIFF
--- a/tests/test_sncosmo_import.py
+++ b/tests/test_sncosmo_import.py
@@ -7,6 +7,7 @@ class sncosmo_import_tests(unittest.TestCase):
         import astropy
     def test_versions(self):
         print(astropy.__version__)
+        self.assert(astropy.__version__=='0')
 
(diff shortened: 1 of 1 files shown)


danielsf:
Given that this is really a test for sncosmo, I think it should be in the sncosmo EUPS package (i.e. our sncosmo repo).

You should just be able to add a ```tests/``` directory to that repository and then copy the ```SConscript``` file from the tests directory of any EUPS package and then proceed as normal.  I did in this in ```sims_maps``` a few months ago to make sure that git-lfs was properly downloading data files.  You can just copy the procedure there.

rbiswas4:
OK.

danielsf:
@rbiswas4 Can I close this PR?

PR #63, opened 2016-07-28
Feature/sim 2068/lc functionality

danielsf:
This pull request does a few things

1) makes it possible to request a limited number of rows from our "high tech" CatalogDBObjects (those that use htmid to do the spatial query)

2) makes it possible to request a limited number of light curves from the LightCurveGenerator

3) moves testSN.py into the tests/ directory so that it will automatically be run by scons.  I had to add a conditional to ensure that the unit tests are skipped on Jenkins, where, for some reason, the environment expected by astropy does not properly load.

### DIFF
--- a/python/lsst/sims/catUtils/utils/LightCurveGenerator.py
+++ b/python/lsst/sims/catUtils/utils/LightCurveGenerator.py
@@ -314,11 +314,14 @@ class LightCurveGenerator(object):
 
         return obs_groups_out
 
-    def _get_query_from_group(self, grp, chunk_size):
+    def _get_query_from_group(self, grp, chunk_size, lc_per_field=None):
         """
         Take a group of ObervationMetaData that all point to the same region
         of the sky.  Query the CatSim database for all of the celestial objects
         in that region, and return it as an iterator over database rows.
+
+        lc_per_field specifies the maximum number of light curves to return
+        per field of view (None implies no constraint).
         """
 
         cat = self._lightCurveCatalogClass(self._catalogdb, obs_metadata=grp[0])
@@ -328,6 +331,7 @@ class LightCurveGenerator(object):
         query_result = cat.db_obj.query_columns(colnames=cat._active_columns,
                                                 obs_metadata=cat.obs_metadata,
                                                 constraint=self._constraint,
+                                                limit=lc_per_field,
                                                 chunk_size=chunk_size)
 
         return query_result
@@ -404,7 +408,7 @@ class LightCurveGenerator(object):
 
             _sed_cache = {}  # before moving on to the next chunk of objects
 
-    def light_curves_from_pointings(self, pointings, chunk_size=100000):
+    def light_curves_from_pointings(self, pointings, chunk_size=100000, lc_per_field=None):
         """
         Generate light curves for all of the objects in a particular region
         of sky in a particular bandpass.
@@ -421,6 +425,10 @@ class LightCurveGenerator(object):
         objects to pull in from the database at a time.  Note: the larger
         this is, the faster the LightCurveGenerator will run, because it
         will be handling more objects in memory at once.
+
+        lc_per_field (optional; default None) is an int specifying the maximum
+        number of light curves to return per field of view (None implies no
+        constraint).
 
         Output:
         -------
@@ -472,7 +480,7 @@ class LightCurveGenerator(object):
             print('starting query')
 
             t_before_query = time.time()
-            query_result = self._get_query_from_group(grp, chunk_size)
+            query_result = self._get_query_from_group(grp, chunk_size, lc_per_field=lc_per_field)
 
             print('query took ', time.time()-t_before_query)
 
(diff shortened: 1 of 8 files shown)


python/lsst/sims/catUtils/utils/LightCurveGenerator.py
```diff
@@ -309,11 +314,14 @@ def get_pointings(self, ra, dec, bandpass=('u', 'g', 'r', 'i', 'z', 'y'), expMJD
 
         return obs_groups_out
 
-    def _get_query_from_group(self, grp, chunk_size):
+    def _get_query_from_group(self, grp, chunk_size, lc_per_field=None):
         """
```
jbkalmbach: For consistency I think we should properly document all input and output parameters here as we have elsewhere.

python/lsst/sims/catUtils/utils/SNIaLightCurveGenerator.py
```diff
@@ -56,6 +57,20 @@ def __init__(self, *args, **kwargs):
         self._brightness_name = 'flux'
         super(SNIaLightCurveGenerator, self).__init__(*args, **kwargs)
 
+    def light_curves_from_pointings(self, pointings, chunk_size=100000, lc_per_field=None):
+        if lc_per_field is not None:
+            warnings.warn("You have set lc_per_field in the SNIaLightCurveGenerator. "
+                          "This will limit the number of candidate galaxies queried from the "
+                          "CatSim database per field-of-view.  Because supernovae are randomly "
+                          "populated in those galaxies, there is no guarantee that the galaxies "
+                          "queried will have supernovae in them.  If the galaxies you actuall "
```
jbkalmbach: typo actually

python/lsst/sims/catUtils/utils/SNIaLightCurveGenerator.py
```diff
@@ -56,6 +57,20 @@ def __init__(self, *args, **kwargs):
...
+                          "query do not host supernovae, you could get fewer light curves than "
+                          "you expect.")
+
```
jbkalmbach: Do we have a way to actually ask and get a certain number of lightcurves with SNe in a pointing? Or plans to?

tests/testAllObjects.py
```diff
@@ -47,6 +48,8 @@ class TestCat(InstanceCatalog):
 
 class basicAccessTest(unittest.TestCase):
 
+    longMessage = True
```
jbkalmbach: Since it seems that pep-8 was fixed in some of the places in this pull request. Shouldn't the variable names in this file be lower_case instead of camelCase?

tests/testAllObjects.py
```diff
@@ -173,6 +176,176 @@ def testObsCat(self):
...
+                # to fatboy, ignore it
+                list_of_failures.append('GalaxyObj')
+                could_connect = False
+            else:
+                raise
+
+        if could_connect:
+            obs_metadata = dbobj.testObservationMetaData
+
+            # query witout a constraint on varParamStr but no limit
```
jbkalmbach: typo, "without"

jbkalmbach:
I couldn't get testAllObjects.py to connect to fatboy even with the ssh tunnel open, but I'm doing this off campus, so just double check that it will actually connect to fatboy with a good connection before merging.

tests/testSN.py
```diff
@@ -328,39 +340,39 @@ class MyGalaxyCatalog(CatalogDBObject):
...
             # columns required to convert the ra, dec values in degrees
             # to radians again
             columns = [('id', 'id', int),
-                       ('raJ2000','raJ2000 * PI()/ 180. '),
-                       ('decJ2000','decJ2000 * PI()/ 180.'),
+                       ('raJ2000', 'raJ2000 * PI()/ 180. '),
+                       ('decJ2000', 'decJ2000 * PI()/ 180.'),
                        ('redshift', 'redshift')]
 
-	# class galCopy(InstanceCatalog):
```
jbkalmbach: What is all this commented out code? Do we need to keep it?

jbkalmbach:
Ok, looks good. Just like I mentioned above make sure that when the tests should be connecting to fatboy that they are.

python/lsst/sims/catUtils/utils/SNIaLightCurveGenerator.py
danielsf: It may not be impossible, but it will take some significant design.  Ordinarily, lc_per_field is implemented as a SQL "LIMIT" on the database query.  For SNIa, we can't actually know how many light curves we got until after all of the processing has been run, so we would have to implement some kind of loop that checks the number of light curves returned and then does another database query (making sure not to get the same objects we got in the last query), repeating until the requested number of light curves have been generated.  Probably, this will need to be done by implementing new versions of most of the LightCurveGenerator methods in SNIaLightCurveGenerator.  I will open a ticket to do so in the future.

The ticket is SIM-2093

tests/testAllObjects.py
danielsf: longMessage is a unittest parameter that, if set to True, means that, if you call ```self.assertEqual(x, y, msg='this is my message')```, then `this is my message' will be appended to the default error message when the assert fails (as opposed to completely replacing the default error message if ```longMessage=False```).

tests/testSN.py
danielsf: I'm not sure.  Rahul would know best.  I erred on the sign of not changing anything I didn't have to when I ported this over.

PR #65, opened 2016-08-10
Feature/sim 1208/flatten

danielsf:


yoachim:
OK, looks like everything runs on all of these.  Might be nice to move the ipynb files over to git-lfs someday, but that's just my preference I think.

merge away.

PR #66, opened 2016-08-31
Feature/sim 2096/phosim interface

danielsf:
A new PhoSim InstanceCatalog API has been published here

https://bitbucket.org/phosim/phosim_release/wiki/Instance%20Catalog

This required a re-working of our CatSim-PhoSim interface.  Under this pull request, whenever a user creates an ObservationMetaData using the ObservationMetaDataGenerator, all of the OpSim columns associated with the pointing will be stored in a dict called `OpsimMetaData`.  The mapping between these OpSim columns and the header parameters expected by PhoSim will be handled by a dict called `phoSimHeaderMap` which is assigned to the InstanceCatalog before writing it.  If no `phoSimHeaderMap` is specified, an error is thrown whose message explains the way the header map should work.  Even if users do not want to map any extra columns (i.e. anything beyond RA, Dec, Alt, Az, MJD, rotSkyPos, and filter, which ObservationMetaData handles automatically), they must specify an empty dict as the `phoSimHeaderMap`.  This is to prevent users from doing something they did not intend (since PhoSim will accept nonsense or incomplete header parameters without failing).

This required changes to

sims_utils
sims_catalogs
sims_catUtils
sims_GalSimInterface


python/lsst/sims/catUtils/baseCatalogModels/OpSim3_61DBObject.py
```diff
@@ -107,15 +104,25 @@ def executeConstrainedQuery(self, spatialBound, constraint=None):
         results = self.execute_arbitrary(query, dtype=dtype)
         return results
 
-    def getObservationMetaData(self, obshistid, radiusDeg, makeCircBounds=True, makeBoxBounds=False):
+    def getObservationMetaData(self, searchCenter, searchRadius, constraint=None,
+                               fovRadius=1.75, makeCircBounds=True, makeBoxBounds=False):
         """
-        This class creates an ObservationMetaData object suitable for generating a PhoSim input
-        catalog based on an actual OpSim pointing.
+        This method will perform a query on the OpSim3_61 for pointings within a given region
```
jbkalmbach: Can we provide a url here in the documentation to where the list of opsim columns can be found? I think that might be useful.

python/lsst/sims/catUtils/exampleCatalogDefinitions/phoSimCatalogExamples.py
```diff
@@ -9,14 +9,69 @@
...
 
-def write_phoSim_header(obs, file_handle):
+# This is a dict of transformations mapping from data in OpSim to header
+# values expected by PhoSim.  The dict is keyed on the names of PhoSim
+# parameters.  The values of the dict are either straight values, in which
+# case those values are written to the PhoSim InstanceCatalog header, or tuples
+# containing the name of the OpSim column which should be use to calculate
+# the PhoSim parameter and any transformation needed to go from OpSim
+# units to PhoSim units.
+DefaultPhoSimHeaderMap = {'rottelpos': ('rotTelPos', np.degrees),
```
jbkalmbach: Maybe a url here to the phosim header information, not sure if it's already elsewhere in the documentation.

python/lsst/sims/catUtils/exampleCatalogDefinitions/phoSimCatalogExamples.py
```diff
@@ -9,14 +9,69 @@
...
 
 
-def write_phoSim_header(obs, file_handle):
+# This is a dict of transformations mapping from data in OpSim to header
+# values expected by PhoSim.  The dict is keyed on the names of PhoSim
+# parameters.  The values of the dict are either straight values, in which
+# case those values are written to the PhoSim InstanceCatalog header, or tuples
+# containing the name of the OpSim column which should be use to calculate
+# the PhoSim parameter and any transformation needed to go from OpSim
+# units to PhoSim units.
```
jbkalmbach: Maybe a url here to the phosim header information, not sure if it's already elsewhere in the documentation.

python/lsst/sims/catUtils/exampleCatalogDefinitions/phoSimCatalogExamples.py
```diff
@@ -9,14 +9,69 @@
...
+        if obs.OpsimMetaData is None:
+            return None
+
+        if transform[0] not in obs.OpsimMetaData:
+            return None
+
+        if transform[1] is None:
+            return obs.OpsimMetaData[transform[0]]
+
+        return transform[1](obs.OpsimMetaData[transform[0]])
```
jbkalmbach: This is cool. I never thought about storing a numpy function like this to use later.

python/lsst/sims/catUtils/utils/ObservationMetaDataGenerator.py
```diff
@@ -70,10 +70,16 @@ def __init__(self, database=None, driver='sqlite', host=None, port=None):
...
+        # a dict keyed on the user interface names of the OpSimdata columns
+        # (i.e. the args to getObservationMetaData).  Returns a tuple that is the
         # (name of data column in OpSim, transformation to go from user interface to OpSim units,
         # dtype in OpSim)
+        #
+        # Note: this dict will contain entries for every column (except propID) in the OpSim
+        # summary table, not just those the ObservationMetaDataGenerator is designed to query
+        # on.  The idea is that ObservationMetaData generated by this class will carry around
+        # records of the values of all of the associated OpSim Summary columns so that users
+        # can pass those values on to PhoSim/other tools and thier own discretion.
```
jbkalmbach: Does this add a lot of excessive memory use? Or is it negligible?

tests/testPhoSimCatalogs.py
```diff
@@ -104,6 +161,226 @@ def testCatalog(self):
         if os.path.exists(catName):
             os.unlink(catName)
 
+    def testHeaderMap(self):
+        """
+        Test the behavior of the phoSimHeaderMap
+        """
+        testBulge = PhoSimCatalogSersic2D(self.bulgeDB, obs_metadata=self.obs_metadata)
+        testBulge.phoSimHeaderMap = {'lunar_distance': ('dist2moon', None),
+                                     'rotation_of_the_telescope': ('rottelpos', np.degrees)}
```
jbkalmbach: Since the default map contians a lambda function, I think we should add a test on an example of that as well.

jbkalmbach:
Looks good.

jbkalmbach:
Oh, just wondering. Has it been a big enough issue with people creating catalogs that give them things they don't want that we don't want a default phosimHeader?

python/lsst/sims/catUtils/baseCatalogModels/OpSim3_61DBObject.py
danielsf: I'd actually rather not, for two reasons:

1) this is a very old OpSim run and I do not think its documentation exists anywhere

2) Because this is implemented as a `CatalogDBObject`, users can use the `get_column_names()` to inspect the available columns.

python/lsst/sims/catUtils/utils/ObservationMetaDataGenerator.py
danielsf: There are only 44 columns in the OpSim Summary table, so I don't think it will be a problem.

danielsf:
I wanted to force the user to set a phoSimHeaderMap (even an empty one) for every PhoSim InstanceCatalog because I am nervous about PhoSim's tendency to run without error even if you pass it garbage.  Given that PhoSim is going to be making decisions for the user about how to handle unset header parameters, I want to force the user to explicitly choose to set no header parameters, if that is really what they want.

If people start complaining to us, we can default the header map to `{}`.

PR #67, opened 2016-09-06
Feature/sim 2108/limited light curves

danielsf:
Modify the LightCurveGenerator so that you can limit the number of light curves drawn form objects based on the GalaxyTileObj CatalogDBObject.


yoachim:
After the usual eups confusion, tests pass and it looks fine to merge.

PR #69, opened 2016-09-21
Twinkles catalog 2

rbiswas4:
Adding changes for using SNDBObj to read Catalogs  of static SN parameters on fatboy.


python/lsst/sims/catUtils/exampleCatalogDefinitions/phoSimCatalogExamples.py
```diff
@@ -290,15 +301,30 @@ class PhoSimCatalogZPoint(PhosimInputBase, PhoSimAstrometryGalaxies, EBVmixin):
...
+class PhoSimCatalogSN(PhoSimCatalogZPoint, FrozenSNCat, EBVmixin):
+    catalog_type = 'phoSim_SNcatalog'
+
+    def get_sedFilepath(self):
+        return self.column_by_name('TsedFilepath')
+
+    def get_phoSimMagNorm(self):
+        return self.column_by_name('TmagNorm')
+
+    default_formats = {'S':'%s', 'f':'%.9g', 'i':'%i'}
```
danielsf: Since you inherit from `PhoSimCatalogZPoint`, you do not need to re-define `default_formats`, `spatial_model`, or `transformations`.

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -1,23 +1,33 @@
 """
-Mixin to InstanceCatalog class to give SN catalogs in catsim
+Mixins for the InstanceCatalog class to provide SN catalogs in catsim. There
```
danielsf: missing word 'are'?

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -1,23 +1,33 @@
...
+three classes here:
+    - SNFunctionality: provides common functions required by all SN instance
+        catalogs. It does not make sense to instantiate this class, but rather
+        it should be used as a mixin alongside another class.
+    - SNIaCatalog: Dynamically created catalogs by sampling user specified
+        distributions of SN parameters on the fly based on host galaxies on the
```
danielsf: 'in the catsim database' rather than 'on the catsim database'?

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -110,33 +125,61 @@ def observedIndices(self):
...
+                        magNorms[i] = np.nan
+                        fnames[i] = None
 
+                # SED in rest frame
+                sed = snobject.SNObjectSourceSED(time=self.mjdobs)
+                try:
+                    magNorms[i] = sed.calcMag(bandpass=bp)
+                except:
+                    # sed.flambda = 1.0e-20 
+                    magNorms[i] = 1000. # sed.calcMag(bandpass=bp)
```
danielsf: Should you also set `fnames[i]=None` in this case?  That way, the PhoSim InstanceCatalog will skip over this supernova, since `sedFilepath` is in the `cannot_be_null` member variable of the PhoSim InstanceCatalog.

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -194,11 +248,14 @@ def get_snbrightness(self):
         if isinstance(bandname, list):
             raise ValueError('bandname expected to be string, but is list\n')
         bandpass = self.lsstBandpassDict[bandname]
+        # Adding photometric parameters
+        _photParams = PhotometricParameters()
```
danielsf: The `PhotometryBase` mixin (and, by extension, `PhotometryStars` and `PhotometryGalaxies`) defines a `self.photParams` member variable that users can change if they see fit.  I am worried that someone might instantiate a galaxy or star catalog with a custom `photParams` and then assign it to a supernova catalog and assume that everything was self consistent.  Can you, instead of creating `_photParams` inside of this method, do something like:

```
if not hasattr(self, 'photParams'):
    self.photParams = PhotometricParameters()
```
and then use `self.photParams` in the rest of the code?

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -1,23 +1,33 @@
...
+        database are the tables `TwinkSN` and `TwinkSNKraken`
 """
 import numpy as np
 
 from lsst.sims.catalogs.definitions import InstanceCatalog
 from lsst.sims.catalogs.decorators import compound
-from lsst.sims.photUtils import BandpassDict
+from lsst.sims.photUtils import (BandpassDict, Bandpass)
 from lsst.sims.catUtils.mixins import CosmologyMixin
 from lsst.sims.catUtils.mixins import PhotometryBase
```
danielsf: It looks like you don't use `PhotometryBase`.  You should remove the import statement.

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -283,6 +345,12 @@ def get_snfluxes(self):
                 vals[:, 16], vals[:, 17], vals[:, 18])
 
 
+    def get_EBV(self):
```
danielsf: Why not just inherit from `EBVmixin` and use the `get_EBV` method there?

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -533,7 +601,79 @@ def get_snparams(self):
...
+
+
+    @compound('snra', 'sndec', 'z', 'vra', 'vdec', 'vr')
+    def get_angularCoordinates(self):
+        '''
+        Obtain the coordinates and velocity of the SN from the host galaxy
+
+        Returns
+        -------
+        `np.ndarray` of coordinara, dec, z, vra, vdec, and vr
```
danielsf: is 'coordinara' a typo?

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -533,7 +601,79 @@ def get_snparams(self):
...
+
+        '''
+        snra, sndec, snz = self.column_by_name('raJ2000'),\
+            self.column_by_name('decJ2000'),\
+            self.column_by_name('Tredshift')
+        snvra = np.zeros(self.numobjs)
+        snvdec = np.zeros(self.numobjs)
```
danielsf: Why define `vra` and `vdec`, if we are automatically populating them with zeros?

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -533,7 +601,79 @@ def get_snparams(self):
...
+
+        return (c, x1, x0, t0)
+
+    def get_snid(self):
+        # Not necessarily unique if the same galaxy hosts two SN
+        # Use refIdCol to access the relevant id column of the dbobj
+        # Should revert to galTileID for galaxyTiled catalogDBObj and
+        # id for galaxyObj catalogDBObj
+        # (email from Scott)
+        return self.column_by_name('Tsnid')
```
danielsf: Should this just return `self.refIdCol`?

The definition of `InstanceCatalog` defines `self.refIdCol = self.db_obj.idColKey`.  Since, in the `SNDBObj`, you defined `idColKey='id'`, this will return the right answer, and it will allow users who want to connect to their own database of supernovae to define a different `idColKey` without having to remember to map it to `Tsnid`.

python/lsst/sims/catUtils/supernovae/snObject.py
```diff
@@ -589,6 +590,94 @@ def SNObjectSED(self, time, wavelen=None, bandpass=None,
         SEDfromSNcosmo.addCCMDust(a_x=ax, b_x=bx, ebv=self.ebvofMW)
         return SEDfromSNcosmo
 
+    def SNObjectSourceSED(self, time, wavelen=None):
```
danielsf: Should this be called `restFramSED()`?  Since it is a part of the class `SNObject`, I do not think it needs 'SNObject' in the method name, and 'restFrame' is what you are actually returning.

python/lsst/sims/catUtils/supernovae/snObject.py
```diff
@@ -589,6 +590,94 @@ def SNObjectSED(self, time, wavelen=None, bandpass=None,
...
+        `numpy.ndarray` of dtype float.
+
+        .. note: The result should usually match the SALT source spectrum.
+        However, it may be different for the following reasons:
+        1. If the time of observation is outside the model range, the values
+            have to be inserted using additional models. Here only one model
+            is currently implemented, where outside the model range the value
+            is set to 0.
+        2. If the wavelengths are beyond the range of the SALT model, the SED
+            flambda values are set to `np.nan` and these are actually set to 0.
```
danielsf: This is a bit confusing.  Should you just say 'flambda values are set to zero', since that is what the user will see?

python/lsst/sims/catUtils/supernovae/snObject.py
```diff
@@ -589,6 +590,94 @@ def SNObjectSED(self, time, wavelen=None, bandpass=None,
...
+        # the ranges of SNCosmo.Model in time. Behavior beyond this is 
+        # determined by self.modelOutSideTemporalRange
+        insidephaseRange = (phase <= source.maxphase())and(phase >= source.minphase())
+        if insidephaseRange:
+            # If SNCosmo is requested a SED value beyond the wavelength range
+            # of model it will crash. Try to prevent that by returning np.nan for
+            # such wavelengths. This will still not help band flux calculations
+            # but helps us get past this stage.
+
+            flambda = flambda * np.nan
```
danielsf: Why set them to Nan?  You've already set everything equal to zero, which is what you want to return in the range outside of the SALT wavelength grid, anyway.

python/lsst/sims/catUtils/exampleCatalogDefinitions/phoSimCatalogExamples.py
```diff
@@ -290,15 +301,30 @@ class PhoSimCatalogZPoint(PhosimInputBase, PhoSimAstrometryGalaxies, EBVmixin):
     transformations = {'raPhoSim': np.degrees, 'decPhoSim': np.degrees}
 
 
+class PhoSimCatalogSN(PhoSimCatalogZPoint, FrozenSNCat, EBVmixin):
```
danielsf: Shouldn't PhoSimCatalogSN set `self.writeSedFile=True` by default?  Is there any case in which a user would want to write a PhoSimCatalog of SNe but not write out the SEDs?

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -32,6 +42,11 @@ class SNFunctionality(object):
     which define these attributes.
     """
 
+    # Write the location of SED file (for example for PhoSim)
+    writeSedFile = False
+    # prefix to use for SED File name
+    prefix = ''
```
danielsf: 'prefix' is too generic a name.  It will get used for something else.  Can this be `sn_sed_file_prefix` or something equally specific?

python/lsst/sims/catUtils/baseCatalogModels/snModels.py
```diff
@@ -0,0 +1,56 @@
...
+import numpy
+from .BaseCatalogModels import BaseCatalogObj
+from lsst.sims.utils import ObservationMetaData
+from lsst.sims.catalogs.db import CompoundCatalogDBObject
+
+__all__ = ['SNDBObj'] 
+
+class SNDBObj(BaseCatalogObj):
+    objid = 'TwinkUnlensedSN'
+    # From now on the tableid should be specified in instantiating the class
```
rbiswas4: There is only one table for other objects on fatboy, so this is specified by default (and perhaps can be changed as an attribute). We need to be able to choose which table to use since we already have two tables and will probably add a third for Twinkles3. 

It seems right to be able to specify this in the constructor.

python/lsst/sims/catUtils/baseCatalogModels/SNModels.py
```diff
@@ -0,0 +1,52 @@
+import numpy
```
rbiswas4: removed

python/lsst/sims/catUtils/baseCatalogModels/snModels.py
```diff
@@ -0,0 +1,56 @@
+"""
+The class SNObj is a catalogDB class which can read a table of SALT2
+parameters on the catsim database
+"""
+import numpy
+from .BaseCatalogModels import BaseCatalogObj
+from lsst.sims.utils import ObservationMetaData
+from lsst.sims.catalogs.db import CompoundCatalogDBObject
```
rbiswas4: removed c7ede749

python/lsst/sims/catUtils/baseCatalogModels/snModels.py
```diff
@@ -0,0 +1,56 @@
...
+               ('Tc', 'c'),
+               ('Tsnid', 'id'),
+               ('Tredshift', 'redshift'),
+               ('Tgaltileid', 'galtileid')
+              ]
+
+    
+    def _final_pass(self, results):
+        """This is to map the values from 0 - 2*PI() as ra goes negative currently"""
+        for ra in ('raJ2000','raJ2000Bulge','raJ2000Disk','raJ2000Agn'):
```
rbiswas4: removed c7ede749

python/lsst/sims/catUtils/exampleCatalogDefinitions/phoSimCatalogExamples.py
rbiswas4: removed cbdc68d

python/lsst/sims/catUtils/exampleCatalogDefinitions/phoSimCatalogExamples.py
rbiswas4: No. changed default: cbdc68d

rbiswas4: Done:  cbdc68d

python/lsst/sims/catUtils/exampleCatalogDefinitions/phoSimCatalogExamples.py
```diff
@@ -45,6 +47,15 @@
                           'seed': ('obsHistID', None)}
 
 
+DefaultPhoSimInstanceCatalogCols = ('object', 'uniqueID', 'RA', 'DEC'\
```
rbiswas4: Done: cbdc68d

python/lsst/sims/catUtils/baseCatalogModels/snModels.py
danielsf: Okay

python/lsst/sims/catUtils/mixins/sncat.py
```diff
@@ -110,33 +125,61 @@ def observedIndices(self):
...
 
-    def get_snid(self):
-        # Not necessarily unique if the same galaxy hosts two SN
-        # Use refIdCol to access the relevant id column of the dbobj
-        # Should revert to galTileID for galaxyTiled catalogDBObj and
-        # id for galaxyObj catalogDBObj
-        # (email from Scott)
-        return self.column_by_name(self.refIdCol)
+    @compound('TsedFilepath', 'TmagNorm')
+    def get_phosimVars(self):
```
rbiswas4: I would like to have the functionality in Instance catalogs. `PhoSimInstance` Catalogs are very specific to phosim because of the headers  are not as useful as methods of carrying over truth information. So, I would think of Instance Catalogs as a vehicle to carry all the information including SEDs and `magnorms` if that is a useful way to think about it.

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: Well, honestly, I thought that this will always work if the previous conditions are satisfied. So, I did not want things to crash if I am missing some case, but don't want it to skip the case completely so that I can look for such occurrances and figure out what is going on.

After a Twinkles is run we can be more confident and aggressive about this if we want.

rbiswas4: But I added a small comment to explain this.

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: Not new code. The idea was people should be free to define these quantities in sncat.

python/lsst/sims/catUtils/supernovae/snObject.py
rbiswas4: I think I prefer the name as is. 

`SNObject` is designed to return a lot of things just the way `SNCosmo` would return them. There are a few differences. The SEDS are one of them. Hence, I think it is important to emphasize the difference to SN users who might be more accustomed to the SNCosmo SED. 
Also, `source` is a specific term that is in use. So I think that is useful to keep.

python/lsst/sims/catUtils/supernovae/snObject.py
rbiswas4: They won't see that if they don't use `self.rectifySED =True`

I agree that this is confusing and a better wording might help.

python/lsst/sims/catUtils/supernovae/snObject.py
rbiswas4: Again not if `self.rectifySED = False`. 

The use cases I am trying to support:
- rectifySED = True : What phosim wants
- rectifySED = False  : Have nans outside the model range to prevent users from using it outside.

python/lsst/sims/catUtils/mixins/sncat.py
danielsf: Okay.  I do think, then, that there's no need to define `TsedFilepath` and `sedFilepath`.  Just have a getter for `sedFilepath` in `sncat.py`.  Ditto for `magNorm`.

python/lsst/sims/catUtils/mixins/sncat.py
danielsf: Okay.

python/lsst/sims/catUtils/mixins/sncat.py
danielsf: One of the things that makes getters (as implemented in InstanceCatalog) so useful is that you can always sub-class something like SNIaCatalog and define a getter for snvra and snvdec.  I'm not sure how that will work now that we have a default getter for them (in fact, I think it won't work; InstanceCatalog will complain if there are two getters for the same column).  If snvra and snvdec are not defined here, then users are free to write their own method to calculate them.  That method can still access snra and sndec from column_by_name().

If you don't want to deal with this now, that is fine.  I think we should look at this question in the future, though.

python/lsst/sims/catUtils/supernovae/snObject.py
danielsf: I'm worried that, if someone does a dir() on SNObject, it is not going to be obvious from the name (SourceSed) that this returns the rest frame SED.

python/lsst/sims/catUtils/supernovae/snObject.py
danielsf: Okay.  I got confused by the '.' after '0'.  I interpreted that as an end-of-sentence period rather than a this-is-a-float decimal point.  Maybe make the '0.' into '0.0'.

python/lsst/sims/catUtils/supernovae/snObject.py
danielsf: Okay

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: yes, fixed

python/lsst/sims/catUtils/supernovae/snObject.py
rbiswas4: Yes, terminology is a problem. a `source` by definition is in rest frame. Maybe we can have an aliase defined too if `restframe` is the language that is more consistent in the catsim universe?

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: OK ... that sounds like I am not appreciating something here. Let us keep this as an extra ticket.

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: Done.

python/lsst/sims/catUtils/supernovae/snObject.py
rbiswas4: Done

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: Actually, a happy byproduct of this, is we can put it in `SNFuncitonality` as this becomes common functionality for now. So done that too.

python/lsst/sims/catUtils/supernovae/snObject.py
rbiswas4: Done that

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: Hmm ... I am not sure why what I have works (or does it)?

The question about using the EBVmixin is what coordinates would be used for a SN, which is tied to a galaxy which are at different positions? In practice this is not much of an issue for what we have now But we already have talked about making `hostless` SN for Twinkles3.

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: Done

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: OK I did make the change for `magNorm`.  `sedFilepath` is harder ... gets into clashes. So leavings that as is.

python/lsst/sims/catUtils/mixins/sncat.py
danielsf: What you have does work, it just seems redundant with code we already have.

The `get_EBV` method on `EBVmixin` bases its E(B-V) calculation on the galactic longitude (`glon`) and galactic latitude (`glat`) of the object.  As long as you provide a way to calculate those quantities for the supernovae, you should be able to use it.

One warning: `EBVmixin` also contains a method `get_galacticAv`, which calculates Av as `EBV*galacticRv`.  That will override anything that is in the database you're connecting to.  I don't know if that is a problem for supernovae or not.

rbiswas4:
Are you saying I can do away with this method altogether as long as a galaxy is assoicated with the SN? 

- When there is no galaxy associated with the SN, how should this work?

python/lsst/sims/catUtils/supernovae/snObject.py
danielsf: If `source` is widely understood to mean "rest frame", then that is fine.  I don't think we should have a separate method that just aliases another.  Go ahead and remove `restFramSED()`.  Sorry for the back-and-forth.

python/lsst/sims/catUtils/mixins/sncat.py
danielsf: Can you point me to the collisions?

danielsf:
I think the only change that I'm still concerned about is `prefix` versus `sn_sed_prefix` (or something equivalent) when writing out supernova SEDs.

python/lsst/sims/catUtils/mixins/sncat.py
rbiswas4: I changed everything back TsedFilepath to get it working. If you want I can try a different branch?

danielsf:
The "galactic" in "galactic longitude" refers to the Milky Way.  It's a coordinate system where glat=0 is in the plane of the galaxy and glon=0 is in the direction of the center of the Milky Way from the Earth

https://en.wikipedia.org/wiki/Galactic_coordinate_system#Galactic_longitude

So: everything has a galactic longitude and galactic latitude

the method `galacticFromEquatorial` in `sims_utils` converts from RA, Dec to glon, glat

python/lsst/sims/catUtils/mixins/sncat.py
danielsf: We can leave TsedFilepath for now.  Let's just try to get rid of it when we get around to adding unit tests for these classes and methods (which will be a separate ticket, as we discussed).

rbiswas4:
Ah .... I think I missed writing the reply to the `prefix` vs `sn_sed_prefix`, but had made the change:
https://github.com/lsst/sims_catUtils/blob/TwinklesCatalog_2/python/lsst/sims/catUtils/mixins/sncat.py#L46

danielsf:
Okay.  Feel free to merge.

rbiswas4:
tests pass with latest commit : https://ci.lsst.codes/job/stack-os-matrix/16312/label=centos-6,python=py2/console

PR #70, opened 2016-10-25
U/danielsf/observation meta data generator on night

danielsf:
In order to be able to use the ObservationMetaDataGenerator in alertsim, we need to be able to

1) query the OpSim database based on the 'night' column
2) query the OpSim database based on an arbitrary, user-specified SQL 'where' clause

This PR adds those capabilities.


danielsf:
It has been pointed out that allowing users to pass in arbitrary SQL constraints is not secure.  I am going to replace this PR with a PR on u/danielsf/obsGen_secure, which adds queries on 'night', but not queries on arbitrary, user-specified SQL where clauses.

PR #72, opened 2016-10-31
U/danielsf/fix phosim extinction

danielsf:
I have

1) fixed the order of extinction parameters in PhoSim InstanceCatalogs
2) added unit tests to test the output of PhoSim InstanceCatalog classes, so we will catch it the next time we break something

There are small supporting PRs in sims_catalogs and sims_GalSimInterface

### DIFF
--- a/python/lsst/sims/catUtils/utils/testUtils.py
+++ b/python/lsst/sims/catUtils/utils/testUtils.py
@@ -36,7 +36,7 @@ def calcADUwrapper(sedName=None, magNorm=None, redshift=None, internalAv=None, i
 
 
 def makePhoSimTestDB(filename='PhoSimTestDatabase.db', size=1000, seedVal=32, radius=0.1,
-                     displacedRA=None, displacedDec=None,
+                     deltaRA=None, deltaDec=None,
                      bandpass='r', m5=None, seeing=None, **kwargs):
     """
     Make a test database to storing cartoon information for the test phoSim input
@@ -62,7 +62,7 @@ def makePhoSimTestDB(filename='PhoSimTestDatabase.db', size=1000, seedVal=32, ra
     @param [in] seeing is the seeing value(s) in arcseconds to be passed to
     ObservationMetaData (optional)
 
-    @param [in] displacedRA/Dec are numpy arrays that indicate where (in relation to the center
+    @param [in] deltaRA/Dec are numpy arrays that indicate where (in relation to the center
     of the field of view) objects should be placed.  These coordinates are in degrees.  Specifying
     either of these paramters will overwrite size.  If you only specify one of these parameters, the other
     will be set randomly.  These parameters are optional.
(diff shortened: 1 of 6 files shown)


danielsf:
@rbiswas4 ping

danielsf:
@rbiswas4 Just wanted to make sure this didn't get lost in all of the Pittsburgh-related madness (also: welcome home).

rbiswas4:
Ah yes, thanks  ... I had forgotten about this. Will finish the review today.

python/lsst/sims/catUtils/utils/testUtils.py
```diff
@@ -1,6 +1,13 @@
...
 from lsst.sims.photUtils import Bandpass, Sed, PhotometricParameters
+from lsst.sims.utils import (_raDecFromAltAz, _getRotSkyPos, _getRotTelPos, Site,
+                             raDecFromAltAz, haversine, ObservationMetaData)
 
-__all__ = ["calcADUwrapper"]
+
+__all__ = ["calcADUwrapper", "makePhoSimTestDB"]
 
 
 def calcADUwrapper(sedName=None, magNorm=None, redshift=None, internalAv=None, internalRv=None,
```
rbiswas4: Method probably deserves a doc string (though it was not changed on this PR)

rbiswas4: Method probably deserves a doc string, though it was not created on this PR, so maybe later.

python/lsst/sims/catUtils/utils/testUtils.py
```diff
@@ -16,13 +23,296 @@ def calcADUwrapper(sedName=None, magNorm=None, redshift=None, internalAv=None, i
...
+    catalog to use.
+
+    The method will return an ObservationMetaData object guaranteed to encompass the
+    objects in this database.
+
+    @param [in] filename is a string indicating the name of the DB file to be created
+
+    @param [in] size is the number of objects int he database
+
+    @param [in] seedVal is the seed passed to the random number generator
```
rbiswas4: It seems that in many places, we are now using `numpy.RandomState` . Is there a reason to use `seed` here?

python/lsst/sims/catUtils/utils/testUtils.py
```diff
@@ -16,13 +23,296 @@ def calcADUwrapper(sedName=None, magNorm=None, redshift=None, internalAv=None, i
...
+    @param [in] bandpass is the bandpas(es) of the observation to be passed to
+    ObservationMetaData (optional)
+
+    @param [in] m5 is the m5 value(s) to be passed to ObservationMetaData
+    (optional)
+
+    @param [in] seeing is the seeing value(s) in arcseconds to be passed to
+    ObservationMetaData (optional)
+
+    @param [in] displacedRA/Dec are numpy arrays that indicate where (in relation to the center
```
rbiswas4: It might be possible to clarify the meaning of this being a deviation from the center rather than the absolute position of the displaced radius if we use something like `deltaRA`. ("in relation to the center" helps, but might be hard to catch)

python/lsst/sims/catUtils/utils/testUtils.py
danielsf: The reason is that this code is much older than my knowledge of `numpy.RandomState` ;)  I will update it.

python/lsst/sims/catUtils/utils/testUtils.py
danielsf: Actually... It looks like this method is not used anywhere.  I will remove it.

PR #73, opened 2016-11-28
fix pupil coordinate getter method

danielsf:
Since we need to account for parallax motion to correctly
get geocentric apparent RA, Dec, the pupil coordinate
getter will go from ra/decObserved to pupil coordinates,
rather than from ra/decICRS (which has no parallax information)


yoachim:
Those unit tests take their sweet time, but looks good to merge.

PR #75, opened 2016-12-15
U/danielsf/matplotlib imports

danielsf:
this should protect those matplotlib imports


python/lsst/sims/catUtils/supernovae/utils.py
```diff
@@ -1,7 +1,16 @@
-import matplotlib.gridspec as gridspec
-import matplotlib.ticker as ticker
-import matplotlib.pyplot as plt
+import warnings
+
+try:
+    import matplotlib.gridspec as gridspec
+    import matplotlib.ticker as ticker
+    import matplotlib.pyplot as plt
+except Exception as ex:
```
danielsf: I don't think we have a convention...  I can use `e`.

PR #78, opened 2017-03-15
Sims py3

danielsf:


python/lsst/sims/catUtils/baseCatalogModels/GalaxyModels.py
```diff
@@ -60,9 +60,9 @@ class GalaxyObj(BaseCatalogObj):
                ('magNormBulge', 'magnorm_bulge'),
                ('magNormDisk', 'magnorm_disk'),
                ('magNormAgn', 'magnorm_agn'),
-               ('sedFilenameBulge', 'sedname_bulge', unicode, 40),
-               ('sedFilenameDisk', 'sedname_disk', unicode, 40),
-               ('sedFilenameAgn', 'sedname_agn', unicode, 40),
+               ('sedFilenameBulge', 'sedname_bulge', str, 40),
```
jbkalmbach: Will these string types work in both python 2 and 3? They were specified unicode before.

python/lsst/sims/catUtils/sersicUtils.py
```diff
@@ -1,100 +0,0 @@
-from scipy.special import gammainc
```
jbkalmbach: Is all this just in the database already? Is that why this is being deleted?

tests/testAllObjects.py
```diff
@@ -106,7 +109,7 @@ def testObjects(self):
 
             # Since there is only one chunck,
```
jbkalmbach: Old code but wanted to note chunk is misspelled here.

tests/testAstrometryMixins.py
```diff
@@ -224,7 +225,7 @@ def testUtilityMethods(self):
                  ('xPix', float), ('yPix', float),
                  ('xFocalPlane', float), ('yFocalPlane', float)]
 
-        baselineData = np.loadtxt(catName, dtype=dtype, delimiter=';')
+        baselineData = np.genfromtxt(catName, dtype=dtype, delimiter=';')
```
jbkalmbach: Is there a reason you switched to genfromtxt here but not in VariabilityMixin.py?

python/lsst/sims/catUtils/baseCatalogModels/GalaxyModels.py
danielsf: In python 3, `str` is unicode-like.  `from builtins import str` imports a python 2 class that mimics that behavior

So: yes.  This will work, and it is still unicode-like.

python/lsst/sims/catUtils/sersicUtils.py
danielsf: I have no idea what this script was supposed to do.  It was about three years old and un-tested (so I'm going to assume it doesn't run).  I have taken this opportunity to `rm` all such code (though I tried to make each `rm` its own commit, in case someone really wants it back).

tests/testAstrometryMixins.py
danielsf: `loadtxt` was being very picky about `dtypes` in python 3.  `genfromtxt` was not.

PR #79, opened 2017-04-17
Sim 2274/mlt flares

danielsf:
This pull request adds the code necessary to use the new MLT dwarf flare variability model ingested into fatboy.  In addition to writing the `applyMLTflaring` method, I had to refactor how `applyVariability` works so that individual `apply*` methods could call `column_by_name` to get secondary data about variable sources (for example, `applyMLTflaring` needs to know the star's parallax; `applyAgn` should need to know a galaxy's redshift, though that has not been implemented, yet).

There is a small supporting pull request in sims_catalog.


python/lsst/sims/catUtils/mixins/VariabilityMixin.py
```diff
@@ -386,56 +488,298 @@ def applyBHMicrolens(self, params, expmjd_in):
...
+        _au_to_parsec = 1.0/206265.0
+        dd = _au_to_parsec/parallax
+
+        # get the area of the sphere through which the star's energy
+        # is radiating to get to us (in cm^2)
+        _cm_per_parsec = 3.08576e16
+        sphere_area = 4.0*numpy.pi*numpy.power(dd*_cm_per_parsec, 2)
+
+        flux_factor = self.photParams.effarea/sphere_area
+
```
jbkalmbach: Ok, this whole process looks right to me and I'm sure you've already done this, but getting another pair of eyes on the last few lines would probably be good.

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
```diff
@@ -386,56 +488,298 @@ def applyBHMicrolens(self, params, expmjd_in):
...
+        lc_name_arr = params['lc'].astype(str)
+        lc_names_unique = numpy.unique(lc_name_arr)
+        for lc_name in lc_names_unique:
+            if 'None' in lc_name:
+                continue
+
+            use_this_lc = numpy.where(numpy.char.find(lc_name_arr, lc_name)==0)
+
+            lc_name = lc_name.replace('.txt', '')
+            if 'late' in lc_name:
```
jbkalmbach: Why is this? Are there 'late_inactive..' in var_params, but no 'late_inactive' light curves in the file?

tests/testMLTflareModel.py
```diff
@@ -0,0 +1,731 @@
...
+                           [17.1, 17.2, 17.3, 17.4],
+                           galacticAvList = [2.432, 1.876, 2.654, 2.364])
+
+        bp_dict = BandpassDict.loadTotalBandpassesFromFiles()
+
+        # calculate the quiescent fluxes of the objects in our catalog
+        baseline_fluxes = bp_dict.fluxListForSedList(sed_list)
+
+        bb_wavelen = np.arange(100.0, 1600.0, 0.1)
+        bb_flambda = blackbody_lambda(bb_wavelen*10.0, 9000.0)
```
jbkalmbach: Nice. I was going to suggest this. Glad you used an outside version as a test. If these are passing I guess my earlier comment about the form of the blackbody must be wrong.

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
```diff
@@ -27,25 +69,52 @@ def reset_agn_lc_cache():
...
+    objects in the base catalog.
+
+    This class provides methods that all variability models rely on.
+    Actual implementations of variability models will be provided by
+    the *VariabilityModels classes.
     """
 
+    _survey_start = 59580.0 # start time of the LSST survey being simulated (MJD)
+
+    # the file wherein light curves for MLT dwarf flares are stored
```
danielsf: good catch

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
```diff
@@ -27,25 +69,52 @@ def reset_agn_lc_cache():
...
+        params_keys = list(params.keys())
+        if len(params_keys) == 0:
+            return 0
+
+        return len(params[params_keys[0]])
+
     def initializeVariability(self, doCache=False):
         """
         It will only be called from applyVariability, and only
         if self.variabilityInitiailized == True (which this method sets)
```
danielsf: You are correct

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
```diff
@@ -59,49 +128,128 @@ def initializeVariability(self, doCache=False):
...
+        # call all of the variability models in the
+        # _methodRegistry to make sure that all of the column
+        # dependencies of the variability models are detected.
+        if len(varParams_arr) == 0:
+            for method_name in self._methodRegistry:
+                self._methodRegistry[method_name]([],{},0)
+
+        # Keep a list of all of the specific variability models
+        # that need to be called.  There is one entry for each
+        # astrophysical object in the CatSim database.
```
danielsf: No.  It is an entry for each astrophysical object.  We ultimately end up running `np.where` on `method_name_arr` to figure out which objects get passed through which methods.  I will try to make the comment clearer.

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
```diff
@@ -59,49 +128,128 @@ def initializeVariability(self, doCache=False):
...
-        """
-        if self.variabilityInitialized == False:
-            self.initializeVariability(doCache=True)
+        expmjd=None
 
-        varCmd = json.loads(varParams)
+        # Loop over all of the variability models that need to be called.
+        # Call each variability model on the astrophysical objects that
+        # require the model.  Add the result to deltaMag.
+        for method_name in numpy.unique(method_name_arr):
```
danielsf: No.  See above.  The `method_name_arr.append()` line occurs outside of the `if varCmd[meth_key] not in method_name_arr` block.

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
```diff
@@ -341,43 +426,60 @@ def applyAmcvn(self, params, expmjd_in):
...
+            zLc[i_burst] += adds
+            yLc[i_burst] += adds
+
+        dMag[0][valid_dexes] += uLc
+        dMag[1][valid_dexes] += gLc
+        dMag[2][valid_dexes] += rLc
+        dMag[3][valid_dexes] += iLc
+        dMag[4][valid_dexes] += zLc
+        dMag[5][valid_dexes] += yLc
         return dMag
```
danielsf: I have run all of the example notebooks we generally give to people.  Everything is fine.  I don't think we ever told the general user base that we were using dicts, so we shouldn't need to worry about people writing their own 'applyVariability' methods that pass out dicts instead of arrays.

Also: our unit test coverage is pretty good, so I think we are fine.

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
```diff
@@ -386,56 +488,298 @@ def applyBHMicrolens(self, params, expmjd_in):
...
+    """
+    A mixin providing the model for cool dwarf stellar flares.
     """
-    This is a mixin which wraps the methods from the class Variability
-    into getters for InstanceCatalogs (specifically, InstanceCatalogs
-    of stars).  Getters in this method should define columns named like
 
-    delta_columnName
+    @register_method('MLT')
+    def applyMlTflaring(self, valid_dexes, params, expmjd):
```
danielsf: Nope.  Twitchy shift-key finger...

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
```diff
@@ -386,56 +488,298 @@ def applyBHMicrolens(self, params, expmjd_in):
...
+
+            ebv_grid = numpy.arange(0.0, 7.01, 0.01)
+            bb_wavelen = numpy.arange(200.0, 1500.0, 0.1)
+            hc_over_k = 1.4387e7  # nm*K
+            temp = 9000.0  # black body temperature in Kelvin
+            exp_arg = hc_over_k/(temp*bb_wavelen)
+            exp_term = 1.0/(numpy.exp(exp_arg) - 1.0)
+            ln_exp_term = numpy.log(exp_term)
+
+            log_bb_flambda = -5.0*numpy.log(bb_wavelen) + ln_exp_term
```
danielsf: Yes, but because I only care about the ratio of flux between the cases where there is and is not dust extinction, I am discarding any normalizing factors.

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
danielsf: This repository

https://github.com/lsst-sims/MW-Flare/tree/master/lsst_sims

Uses the same algorithms I used here.  The results are validated against various sources of real data.  The magnitude ranges all appear to agree with the data.  The README in the lsst_sims directory explains where there are scripts to generate plots illustrating my argument.

python/lsst/sims/catUtils/mixins/VariabilityMixin.py
danielsf: There isn't supposed to be a 'late_inactive' class, but I was stupid an assigned some 'late_inactive' stars n fatboy.  Rather than fix the fatboy curves (which will take about a week of compute time), I figured I'd fix it here.

PR #80, opened 2017-05-31
Attempt to fix problem by adding parameter to method catsimBandMag

rbiswas4:
- The entire problem stems from having to calculate magnitudes. We
  introduce an extra parameter `noNan` in
  ``` catsimBandMag(self, bandpassobject, time, fluxinMaggies=None,
                    noNan=False)
  ```
  so that for evaluation of magnitudes (which is done in the Instance
  Catalogs), this continues to produce `np.nan` which is avoided by
  the instance catalogs. However, in the evaluation of
  `catsimBandFluxError` and `catsimBandMagError` this function is
  called with `noNan = True`. In this case the magnitude is evaluated
  to be 200.0 when the actual calculation would lead to a `np.nan`
  ie. cases where the flux is 0. or negative. This is a super faint
  magnitude and we should get the correct result of not using the
  source photons
	modified:   snObject.py

- While we had tried with a value of mag=1000, this was found to have problems. The main problem was that SNR was going to 0. with a default mag of 1000. due to underflows.  Changed it to 200.

- Check that error calculated is a float and is not nan/infinity
	modified:   tests/testSN.py

- Check that we get approximately sky noise
	modified:   examples/SNObject_example.ipynb


danielsf:
There are just some changes I would make to docstrings and comments.  Otherwise, this looks good to merge.

danielsf:
@rbiswas4 Are you going to merge this?

rbiswas4:
Yes, sorry. I will do this over the next couple of days.

PR #81, opened 2017-06-09
U/danielsf/debug/mba validation

danielsf:
This pull request just adds some functionality that I noticed was missing while I was trying to validate our asteroid models.  It adds a mixin that calls the `chipNameFromPupilCoordsLSST()` method, which is about a factor of six faster than `chipNameFromPupilCoords(camera=LsstSimMapper().camera)`.


yoachim:
Looks good. 

On an unrelated note, `testSN.py` failed for me because pandas depreciated `DataFrame.sort`, so update line 470 to use `.sort_values()`

PR #83, opened 2017-06-23
U/danielsf/debug/profile lightcurve

danielsf:
This pull request should implement a faster version of the LightCurveGenerator for cases where the variability model is simply

quiescent_mag + delta_mag(t)

where quiescent_mag does not vary as a function of time (this is in contrast to the SNIa variability model, which is a full spectro-temporal variability model in which, at each time step, a new supernova SED must be created and integrated over the LSST bandpasses).

If this branch, as it exists, still is not fast enough for alertsim, we can iterate further before closing and merging this pull request.


danielsf:
@sambolino Are you happy enough with this for me to merge it?

sambolino:
Good to merge. Tx.

PR #84, opened 2017-07-28
Sim 2408/nfs bug

danielsf:
Some unit tests were failing on NFS systems because of open database connections left dangling after the tests were run.  This pull request modifies those tests to explicitly call `sims_clean_up()`, closing those connections before running the standard LSST memory leak tests.

### DIFF
--- a/python/lsst/sims/catUtils/mixins/VariabilityMixin.py
+++ b/python/lsst/sims/catUtils/mixins/VariabilityMixin.py
@@ -661,6 +661,8 @@ class MLTflaringMixin(Variability):
             _MLT_LC_NPZ = numpy.load(self._mlt_lc_file)
             sims_clean_up.targets.append(_MLT_LC_NPZ)
             _MLT_LC_NPZ_NAME = self._mlt_lc_file
+            _MLT_LC_TIME_CACHE = {}
+            _MLT_LC_FLUX_CACHE = {}
 
         if not hasattr(self, '_mlt_dust_lookup'):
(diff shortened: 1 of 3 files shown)


python/lsst/sims/catUtils/mixins/VariabilityMixin.py
```diff
@@ -640,7 +641,7 @@ def applyMLTflaring(self, valid_dexes, params, expmjd,
                                "knowledge of the effective area of the LSST "
                                "mirror.")
 
-        if _MLT_LC_NPZ is None or _MLT_LC_NPZ_NAME != self._mlt_lc_file:
+        if _MLT_LC_NPZ is None or _MLT_LC_NPZ_NAME != self._mlt_lc_file or _MLT_LC_NPZ.fid is None:
```
cwwalter: I'm not sure how the fid attribute gets set below and also why it would be necessary to check for it since if you load the numpy file the first two items in this list will be set.  What is different about the fileId?

danielsf: I check for the fid attribute because, if someone runs sims_clean_up(), and then tries to simulate variability again, _MLT_LC_NPZ will not be `None`, but it will be closed (in which case, _MLT_LC_NPZ.fid will be `None`).

danielsf:
I named `_MLT_LC_NPZ` because the light curves are stored as a 'numpy zip' file (with suffix `.npz`) created with `numpy.savez()`.

PR #85, opened 2017-08-20
SIM-2425: Work with pytest-xdist

timj:
This is a work in progress. Not ready for release.

The main thrust is to use temporary file names and temporary scratch space so that multiple processes can run tests from the same class. This work is similar to that in lsst/sims_catalogs#13

### DIFF
--- a/examples/baseValidation/utils.py
+++ b/examples/baseValidation/utils.py
@@ -129,7 +129,7 @@ def get_TotalSDSSMags(result, bandpasses=('u','g','r','i','z')):
     datadir = os.environ.get("SIMS_SED_LIBRARY_DIR")
     tpath = os.getenv('SDSS_THROUGHPUTS')
     bands = {"u":None, "g":None, "r":None, "i":None, "z":None}
-    for k in bands.keys():
+    for k in bands:
         bands[k] = Bandpass()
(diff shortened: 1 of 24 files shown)


timj:
This all seems to be working now (I managed to do a `pytest -n 4 tests/tests*.py`).

rbiswas4:
I actually was only reviewing the parts relating to SN and had tried to approve those changes. I actually have not reviewed all of it (thinking that Scott might have other thoughts, but this shows up as an approved review for the entire PR.

timj:
Thank @rbiswas4 -- I've added @danielsf as reviewer.

timj:
Yes. I deliberately used `mktemp` because I didn't want to have to delete the file that `mkstemp` created. I also thought the chances of a random clash during testing were low and I don't think that anyone is going to try to guess the file as a way to attack the process running the tests. If you are going to insist on `mkstemp` I'll switch.

danielsf:
No need to change.  I just wanted to make sure you were aware.

I just got a failure of `testFastLightCurveGenerator.py` on my local system.  Let me diagnose that, then I will mark this as 'approved'.

danielsf:
This is ready to merge (there was a block of mis-indented code that was causing a failure in python 2 but not python 3)

timj:
Thanks. Python really doesn't like big blocks where it's hard to get the indent right...

PR #88, opened 2017-08-28
do not check for existence of camera until actually writing catalog

danielsf:
this will allow users to add a camera object after catalog
instantiation


danielsf:
I'm going to have to merge these changes early to deal with the API changes made in DM-5922

PR #89, opened 2017-09-20
Sim 2303/kplr lightcurves

danielsf:
This pull requests adds the mixin which generates stellar variability by loading a parametrized light curve function and evaluating it at simulated observation times.  Light curves are parametrized as the sums of sinusoids.  This mixin is meant to be used with the parametrized fit to Kepler light curves derived from

https://github.com/danielsf/CatSimLC

### DIFF
--- a/python/lsst/sims/catUtils/mixins/VariabilityMixin.py
+++ b/python/lsst/sims/catUtils/mixins/VariabilityMixin.py
@@ -830,7 +830,48 @@ class MLTflaringMixin(Variability):
 class ParametrizedLightCurveMixin(Variability):
     """
     This mixin models variability using parametrized functions fit
-    light curves.
+    to light curves.
+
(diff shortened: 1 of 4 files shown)


yoachim:
oh, you're going to love this one. Looks like in python 3 gzip.py wants bytes not str in all the write commands. 

I made a pass at patching it up.

* changed the gzip to write bytes
* had the read method check if a line is bytes and convert it back to a str
* Added some checking because `np.unique` barfed if there were `None`s.
* wrapped some values in `np.max` because they were 1-element arrays and couldn't be compared to floats.

danielsf:
I changed `_calc_dflux` so that it casts the output to a `float` if it is a 1-element numpy array.  That will probably be less aggravating for users who don't want to have to append `[0]` to everything.

PR #90, opened 2017-10-11
U/danielsf/debug/connection tests

danielsf:
Several small changes have left us in a state where, if you can actually connect to fatboy and run the tests of those connections, you will get unit test failures.  We did not notice this because Jenkins cannot connect to fatboy and automatically skips the offending tests.  Unfortunately, the alertSim team develops on a whitelisted machine which can connect to fatboy, meaning that installing/updating the stack triggers those tests and causes failures.  This PR fixes those tests.  In order to exercise the changes, you will need to run the unit tests while a tunnel to fatboy is open (it is not enough to be on campus and thus have access to fatboy directly; the tests assume you are using the ssh tunnel through port 51433).

I am sending you this pull request because some of the changes required touching the moving object sims, updating them to reflect our updated MBA tables on fatboy.

### DIFF
--- a/python/lsst/sims/catUtils/baseCatalogModels/SsmModels.py
+++ b/python/lsst/sims/catUtils/baseCatalogModels/SsmModels.py
@@ -183,6 +183,13 @@ class MBAObj(SolarSystemObj):
     """
     ssmtable = 'fSSMMBABase'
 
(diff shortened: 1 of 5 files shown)


rhiannonlynne:
Looks fine to me.

PR #94, opened 2018-02-14
U/danielsf/factor out phosim

danielsf:
This PR implements some changes needed to support a refactor of ImSim.

### DIFF
--- a/python/lsst/sims/catUtils/exampleCatalogDefinitions/phoSimCatalogExamples.py
+++ b/python/lsst/sims/catUtils/exampleCatalogDefinitions/phoSimCatalogExamples.py
@@ -266,7 +266,7 @@ class PhosimInputBase(InstanceCatalog):
         write_phoSim_header(self.obs_metadata, file_handle, self.phoSimHeaderMap)
 
 
-class PhoSimCatalogPoint(PhosimInputBase, PhoSimAstrometryStars, EBVmixin):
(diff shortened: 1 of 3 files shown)


python/lsst/sims/catUtils/exampleCatalogDefinitions/phoSimCatalogExamples.py
```diff
@@ -266,7 +266,7 @@ def write_header(self, file_handle):
         write_phoSim_header(self.obs_metadata, file_handle, self.phoSimHeaderMap)
 
 
-class PhoSimCatalogPoint(PhosimInputBase, PhoSimAstrometryStars, EBVmixin):
```
jbkalmbach: Why is this done? And why is it okay to do it here, but it's not necessary in `PhoSimCatalogZPoint` or `PhoSimCatalogSN`?

danielsf: `EBVmixin` provides a `get_EBV` method that returns E(B-V) interpolated from a 2-D map (so: including the entire depth of the Milky Way).  For extragalactic sources, that is what we want.  For stars, we want E(B-V) to come from the 3-D dust model stored on fatboy.  The way CatSim works, the presence of getter methods overrides data from fatboy.

PR #98, opened 2018-07-09
remove import of pyfits

danielsf:


danielsf:
The import of `pyfits` wasn't even used, and the resent rework of the `lsstsw` environment broke `sims_catUtils`, anyway, so I'm just going to merge this.

PR #99, opened 2018-07-10
U/yoachim/move map making over

yoachim:
Stuff that got deleted from MAF that needs a place to live for posterity. The stellar density is probably bit rotted.


yoachim:
Good point. But they are kind of more than examples. I'll toss them in a `bin.src` directory, I think that makes the most sense.

PR #100, opened 2018-09-28
U/danielsf/thread/agn

danielsf:
Make it possible to multithread the AGN random walk

This will help with DC2 InstanceCatalog generation

I haven't added any unit tests (I'm not sure if it is fair to assume that a system building this code will have more than one thread to use).  Feel free to suggest tests that should be added.

(There's probably also a lot of chaff lying around that I used in developing this; I apologize for that)


danielsf:
Good idea.  I'll add that unit test before merging.

PR #101, opened 2018-10-15
U/danielsf/opsim/v4

danielsf:
There is a user who is simulating light curves of tidal discruption events using the `LightCurveGenerator` class defined in this package.  She wanted to be able to use the OpSim v4 simulations.  This PR adds a subclass of `ObservationMetaDataGenerator` that can parse OpSim v4, and allows that class to be used in the `LightCurveGenerator`.

This PR does not deal with all of the issues we will have to fix given the new OpSim database schema (PhoSim InstanceCatalog headers will have wonky units and missing header fields), but I wanted to get this out the door so that user wasn't stuck twiddling her fingers longer than needed.  We will deal with the PhoSim interface once DESC starts thinking about DC3 image simulation (or once someone asks for it).

There is also a pull request in `sims_catalogs`.


rhiannonlynne:
I'd like to suggest that you modify ObservationMetadataGenerator a bit -- You could make a generic ObservationMetaDataGenerator "constructor" that has a standard interface for the user, but then calls either v3 or v4 (i.e. you make ObservationMetaDataGeneratorV3 and ObservationMetaDataGeneratorV4), and the generic one calls the appropriate class and returns that. 

This is what we ended up doing in MAF to support v3 and v4 databases, without having to make it super obvious to the user (and modifying it in tons of scripts) -- see https://github.com/lsst/sims_maf/blob/master/python/lsst/sims/maf/db/opsimDatabase.py

I think it will save you time modifying code later, saves you making the "is it v3 or v4" in every script, and makes it easier for the user (especially for backwards compatibility in their or your code).

rhiannonlynne:
Or alternatively, since the ObservationMetadataGeneratorV4 class is so thin (basically just changing the dictionary sent into the init and the name of the table for the summary?) maybe you could move the dictionaries into a supporting class or function (which is just a dictionary holder .. like https://github.com/lsst/sims_maf/blob/master/python/lsst/sims/maf/batches/colMapDict.py maybe?) and just add a method to retrieve the right one and send it in. (and you could add the summary table name to the dictionary, then grab it from there).

danielsf:
Okay.  There is now one `ObservationMetaDataGenerator` class that can work with both v3 and v4 OpSim databases.  I had added an `opsim_version` element to `ObservationMetaData.OpsimMetaData` to try to save us from generating v3 PhoSim InstanceCatalogs with v4 `ObservationMetaData`s and vice versa.

danielsf:
@rhiannonlynne I know you were busy with scheduler meetings this week.  I just wanted to ping you because I *think* I have addressed your concerns on the original PR.

Thanks.

rhiannonlynne:
Thanks @danielsf! Looks good to me.

PR #104, opened 2019-10-21
U/danielsf/tiles

danielsf:
This is the pull request in support of moving the CatSim server from the Windows machine over to epyc. Here are the instructions for actually connecting remotely to the database on epyc (I will slack you the password and username).

The ssh command to establish the tunnel to the server is now:

ssh -N -L 51433:localhost:1433 lsst-sims@epyc.astro.washington.edu

The new CatalogDBObject classes you want to use are:

StarObj -> UWStarCatalogObj
GalaxyTileObj -> UWGalaxyTileObj
GalaxyDiskObj -> UWGalaxyDiskObj
GalaxyBulgeObj -> UWGalaxyBulgeObj
GalaxyAgnObj -> UWGalaxyAgnObj

all of which can be imported from lsst.sims.catUtils.baseCatalogModels


data/tile_data.txt
```diff
@@ -0,0 +1,4051 @@
+#i;ra;ec;corners
```
jbkalmbach: Should this be labelled 'dec' instead of 'ec'?

tests/testLocalGalaxyTileObj.py
```diff
@@ -0,0 +1,463 @@
...
+
+import lsst.utils.tests
+from lsst.utils import getPackageDir
+
+import lsst.sims.utils.htmModule as htm
+from lsst.sims.utils import xyz_from_ra_dec
+from lsst.sims.utils import ra_dec_from_xyz
+from lsst.sims.utils import angularSeparation
+from lsst.sims.utils import ObservationMetaData
+from lsst.sims.catUtils.baseCatalogModels import LocalGalaxyModels as LocGal
```
jbkalmbach: I wasn't able to import this and couldn't find it. Should this be UWGalaxyModels?

danielsf: Yeah.  I changed the name of the classes before showing anyone else this code. Forgot to `grep` the unit tests.

danielsf:
The `workspace/alertsim_190329` code is my latest attempt to figure out how best to simulate the full LSST alert stream.

It was run back in the spring when Zeljko was asking for such a simulation. It hasn't been run since.

I'm not sure if this is yet the best approach to alert simulation. I think it might be valuable to have (unless it is so inscrutable that the next person to try can't really make use of it). What was your reaction? Did it make sense?

jbkalmbach:
I think so. From what I can tell it looks like you are running through the catalogs for a set of observations and then for each type of variable class you calculate the magnitude for the objects (subtracting the quiescent magnitude for AGN) in the observation specified and are checking if that goes above a noise threshold and cataloging the results that do. It makes sense, just a lot of machinery involved, haha.

danielsf:
Yeah. The trickiest part is efficiently figuring out which observations even have a chance of detecting which sources so that you can efficiently generate the light curves.

It sounds like you were able to understand the workflow, which is a relief.

PR #106, opened 2020-02-12
Prevent None files being produced when creating SN instance catalogs.

jbkalmbach:
I think this quick fix prevents the `None` files with an empty SED from appearing when running instance catalog generation. The problem seemed to be that SNe outside the time range were still writing instance catalogs.


jbkalmbach:
@rbiswas4 If you just want to check that this doesn't affect any intended functionality while solving the problem that'd be good. Thanks.

jbkalmbach:
Great, if nobody has any objections I will merge this at 4pm Pacific this afternoon.

PR #108, opened 2020-03-03
Disable tests in tests/testSN.py.

jdswinbank:
Due to incompatibility with latest SNCosmo release.


rhiannonlynne:
@rbiswas4 has provided an update to the failing unit test that should fix this problem.

PR #109, opened 2020-03-04
Fix failures in `testSN.py` following astropy and sncosmo updates

rbiswas4:
-  Added relative tolerances to tests so that comparisons are not
required to be insanely accurate

- : Fix failure due to wavelengths outside range.

The sncosmo model being used is salt2-extended which has a lower
wavelength range of 1700 Ang. The test required evaluating the
`sncosmo` spectrum for an object at z = 0.96 at a wavelength of 3000
Ang, which corresponds to a rest frame wavelength of 1530.6 Ang. To fix
this problem the self.wave attribute of the test was changed to start
form a lowest wavelength of 3500 Ang.

NOTE: the failure of the test was due to comparing directly with a
`sncosmo` model. `SNObject` models are inensitive to this (and provide
 values of 0 if the spectra are not available, with this being oe of the
 desired changed from `sncosmo`. This would not fly for
cosmology but is better to have rather than end up with failures in
simulation.
	modified:   testSN.py

-  fixing deprecated import for comparison of pandas dataframes
triggering warning
	modified:   testSN.py

### DIFF
--- a/tests/testSN.py
+++ b/tests/testSN.py
@@ -26,7 +26,7 @@ import shutil
 
 # External packages used
 # import pandas as pd
-from pandas.util.testing import assert_frame_equal
+from pandas.testing import assert_frame_equal
 import sncosmo
 import astropy
 
@@ -88,7 +88,7 @@ class SNObject_tests(unittest.TestCase):
         print('===============================')
         print('===============================')
         # A range of wavelengths in Ang
-        self.wave = np.arange(3000., 12000., 50.)
+        self.wave = np.arange(3500., 12000., 50.)
         # Equivalent wavelenths in nm
         self.wavenm = self.wave / 10.
         # Time to be used as Peak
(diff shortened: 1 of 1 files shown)


rhiannonlynne:
Just to clarify - are you saying that this would not have been a failure for sims_catUtils, as it would have already accounted for this? (but that you've modified what we're asking for from sncosmo directly to avoid the unit test failure)

rbiswas4:
Exactly ... the priority in `sims_catUtils` was to continue working with spectra outside the range of validity of the model assuming the spectrum was zero (it is expected to be small) outside the range, while the priority in sncosmo was to inform the user by throwing an error so that a cosmological analysis would not be wrong. `sims_catUtils` continues to work correctly. 

To ensure that we continue to check that the `sims_catUtils` sed works beyond the range, I modified the test. Now it evaluates the `sims_catUtils` sed over the old wavelength range (and essentially checks that no errors are thrown), but obtains the `sncosmo` sed over a shorter wavelength staying in the model range, and does the comparison of the two seds over the common range in wavelength. This achieves both of what I had yesterday, but also tests that we will not throw an error outside the range of the model.

rhiannonlynne:
u/rbiswas4/sims_catUtils_updates branch deleted.

